PR #29, opened 2024-11-12
Overlap checker tool

Waqar-ukaea:
Added overlap checker tool from DAGMC to check and identify overlapped regions in .h5m faceted files.

- Checks for overlaps by looping through every triangle vertex in a mesh. 
- Added as a new tool 
- Cannot sample additional points along triangle edges (like DAGMC's version) as `XDG::MeshManager` does not have support to call MOAB functions that return entity adjacencies 

Tested the tool on some h5m files locally and it seems like its working okay. It identifies when there are overlaps and which volumes are overlapped in agreement with DAGMC's version. Also added a new test file that runs the 5 test cases that DAGMC does for its overlap-checker tool. 

One thing to note, the overlap location printed when running the tool doesn't necessarily match DAGMC's. I'm not sure what the cause of that is or if that is a major concern? Maybe due to differences in the order in which vertices are checked between the libraries?

Edit - CI fails because the .h5m files needed for those tests are missing from the repository. The files to be used with overlap-check's test's can be found here - [test_files.zip](https://github.com/user-attachments/files/17731916/test_files.zip). The CMakeLists.txt expects these files to be placed in `${XDG_ROOT}/tests/test_files`


tests/test_overlap_check.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+TEST_CASE("Overlapping Volumes Test")
+{
+  // Create a mesh manager
+	std::shared_ptr<XDG> xdg = XDG::create(MeshLibrary::MOAB);
+	const auto& mm = xdg->mesh_manager();
+
+	mm->load_file("overlap.h5m");
+	mm->init();
+  xdg->prepare_raytracer();
```
pshriwise, line 22: The indentation got a little off here it looks like

tests/test_overlap_check.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+TEST_CASE("Non-Overlapping Volumes Test")
+{
+  // Create a mesh manager
+	std::shared_ptr<XDG> xdg = XDG::create(MeshLibrary::MOAB);
+	const auto& mm = xdg->mesh_manager();
+
+	mm->load_file("no_overlap.h5m");
+	mm->init();
```
pshriwise, line 41: Same as above. Indentation looks off.

pshriwise (changes requested):
CI looks good. All set to be merged after the very minor changes requested here.

Waqar-ukaea:
Fixed the inconsistent indentations you highlighted and a couple more I noticed, lmk if anything else needs editing.

I also removed the parsing of an argument to hold the number of points along a triangle edge to sample since I don't think we will be implementing that feature anymore and will instead go straight to ray_fire along edges check.

pshriwise:
@Waqar-ukaea it looks like CI is still unhappy. Did the submodule get updated after that PR to the test files repo?

PR #31, opened 2024-12-04
libMesh backend

pshriwise:
**This PR description will be updated as the changes evolve to be more fully-featured.**

A backend for libMesh tetrahedral meshes. This is pretty experimental at the moment, but the initial tests indicate that ray tracing on the boundary of tetrahedral mesh subdomains is functional.


pshriwise:
Before marking this as ready, I'm going to summarize the capabilities introduced so what's left to be implemented is well understood and any unused code/methods can be tabled for the next implementation.

pshriwise:
After the addition of the latest test case and a refactor of a larger method on the `MeshManager` class, I'm pretty happy with this at this point. I'd like to add the latest test case to the benchmarking repo, but then I think this will be good to go as an initial version.

There's probably some renaming to consider as well -- `LibMeshMeshManager` sounds too akin to `Boaty McBoatFace` to me.

pshriwise:
Okay I think this is ready to go in. It's been tested on a few models, and (with some manual thread settings) its in the process of being benchmarked for transport using MOAB and libMesh models with identical tessellations.

PR #35, opened 2024-12-18
Improved overlap checker

Waqar-ukaea:
Draft PR to add an algorithm for checking overlaps along geometry edges to the overlap checker tool. 

Current state:
- Loops over all triangles and fires rays from every triangle vertex fires a ray along the edge checking for an intersection with a surface belonging to a volume that is **not** the parent volume of the triangle. This works and correctly identifies intersections where I would expect. However...
- There are additional intersections detected from rays that are launched from one volume and collide with another volume (as they are not stopping at the boundary of the volume they are launched from) leading to intersections on the other volume (see image below). 
- Perhaps the extra intersections are inevitable due to the way rays travelling along surfaces are handled but I would need a way to isolate the intersection points of interest (where the edges overlap) - not sure how I would do this just yet.
Screenshot from 2024-12-18 16-50-05
- Explicitly refers to elements as triangles - should probably change variable names to reflect generic elements. In theory this shouldn't require any logic changes just nomenclature.
- By default the overlap-check tool will only check for overlaps on vertices but edge checking can be enabled by passing flags `-e` or `--check-edges` at runtime.
- Lots of extra fluff in writing things out to files and stdout which needs to be removed before merging into main.
- Missing a progress bar.
- Missing support for openmp parallelism over the new algorithm.
- Added skeleton for an additional test in the overlap checker tests. This currently doesn't test for anything but is in place ready for once I am happy with the edge algorithm for the overlap checker. The test relies on the geometry added in [PR#3](https://github.com/pshriwise/xdg_test_files/pull/3#issue-2747428190) for the [xdg_test_files](https://github.com/pshriwise/xdg_test_files) repo.


pshriwise:
I updated the test files submodule here and included the symbolic link line for the test in the CMakeLists file. I'll do a more thorough review soon if you feel like this is ready to go @Waqar-ukaea

Waqar-ukaea:
I think I may have accidentally overwritten the changes you made the tests which has the tests failing again. I'll fix them in my next commit.

Waqar-ukaea:
**PR status changed to ready for review**

Changes to `xdg::ray_fire()` (namely the addition of a parameter for max distance a ray travels) has fixed the issues I was having previously. Also, refactoring the main loop to loop over all the surfaces has made it easier to identify the volumes which don't need to have rays fired against. Now loops over all surfaces, over all elements on that surface and all edges in each element, firing a ray against all other volumes (excluding the forward/reverse sense parents of the surface).

Tested on a couple geometries and it looks to be working as expected identifying overlaps where I would expect them to be.

Tasks from previous comment:
- [x] Tests implemented
- [x] OpenMP Parallelism over main loop (granted I haven't profiled it)
- [x] Progress Bar
- [x] Consistent results reporting with vertex checker. Maybe we actually want to report more information on this front?
- [x] Code cleaned up/refactored
- [x] Options to write out extra data to text files for plotting in Paraview etc:
   - Ray Paths of those which meet overlaps 
   - Collision locations of these overlaps
   - Ray launch positions and directions
- [x] Consistent results reporting with vertex checker. Maybe we actually want to report more information on this front?
- [x] Code cleaned up/refactored

Waqar-ukaea:
As per the conversation we had on Friday. I have removed the extra file IO components I used for debugging and have replaced them with a new runtime flag --verbose or -v to enable more verbose outputs to stdout. The information being output is also much simpler than I previously had, being limited to just the locations in xyz coords of every edge ray_fire() collision detected along an edge, when the -ve flag is enabled.

I could also extend this to report the locations of every vertex overlap when the flag is enabled too, if that would be of interest.

pshriwise:
> I could also extend this to report the locations of every vertex overlap when the flag is enabled too, if that would be of interest.

I think that would be good if you're up for it!

include/xdg/overlap.h
```diff
@@ -10,9 +13,32 @@
...
+std::vector<EdgeRayQuery> return_ray_queries(const ElementVertices &tri);
 
-void report_overlaps(const OverlapMap& overlap_map);
\ No newline at end of file
+MeshID check_along_edge(std::shared_ptr<XDG> xdg, 
+                       std::shared_ptr<MeshManager> mm, 
+                       const EdgeRayQuery& rayquery, 
+                       const std::vector<MeshID>& volsToCheck, 
+                       std::vector<Position>& edgeOverlapLocs);
+#endif
```
pshriwise, line 44:
```suggestion
#endif // include guard
```

pshriwise (changes requested):
Really just small line comments from me here @Waqar-ukaea. Thank you for the nice work!

As you suggested I do think that adding the verbose output for vertex checks would be good.

Also, let's perform the edge checks by default. If someone is running this I think their expectation is that it will be a comprehensive search by default.

Waqar-ukaea:
Resolved all of the comments (I agree with all with them) and made some additional changes as requested. 

- Removed `calculate_direction()` and `calculate_distance()` functions from `overlap.cpp` and `overlap.h`. I figured these would probably be redundant when I wrote them but couldn't figure out another way at the time. 
- Edge checking is now enabled by default and the `-e` flag now disables edge tracking (full flag name `--disable-edge-tracking`.
- Verbose flag now reports vertex overlap locations as well as edges. I kind of want to refactor the entire function which does the vertex checking but I am going to leave it for now because it does work. Technically this isn't reporting the overlap location but rather the position which returns `true` from `xdg::point_in_volume()`.
- Fixed some more inconsistent indentations I spotted. On this point, for whatever reason the indentations look fine when I'm editing in vscode but when opening pushing to github (or opening in another editor locally) I notice the indentations are off. If I have time I'll look into fixing that and/or setting up a linter to avoid this happening in the future.

But otherwise I think its ready to be merged :)

pshriwise:
@Waqar-ukaea it appears that the version of Embree in CI was updated and revealed an issue with CMake discovery of that package in XDG. I've corrected it in #44 but you'll need to pull those changes in here to get CI working agin.

Waqar-ukaea:
@pshriwise CI fix merged in and I've updated the test_files submodule to include the new test files because it was still failing somehow. Either way CI now passes.

pshriwise (approved):
Thank you @Waqar-ukaea!

PR #41, opened 2024-12-25
Cache build dependencies to speed up CI

pshriwise:
This caches the MOAB build to speed up CI. I'll apply it to other dependencies as they come up.


pshriwise:
Time for OpenMC testing is cut by 20 min here 👍🏻

PR #46, opened 2025-01-28
Bare bones documentation

pshriwise:
This lays out a file tree and some really barebones documentation for XDG.


pshriwise:
Note to future: Squashed all commits here to ensure some of the logos I was considering aren't in the commit history.

PR #57, opened 2025-02-10
Resolve #4 to remove the extra triangle references being stored

### LINKED ISSUE #4: Upgrade to remove triangle reference duplication
Right now there are two triangle references per triangle: one with a forward sense in its scene (volume) and another for the reverse sense in its scene (another volume).

We may be able to avoid the need to duplicate these references by storing the forward and reverse sense scenes on the `GeometryData` object -- it should be available to us as part of the intersection arguments. The remaining need is to know what scene the ray is being fired on. This information can be stored on the `RTCDRay` if needed, but it may also be available as part of the intersection arguments. TBD.

Waqar-ukaea:
Wanted to get a draft PR up to resolve #4. 

- Added a `MeshID` for forwards and reverse sense to the `GeometryUserData` struct
- Replaced previous references to `primtive_ref.sense` to use the new members in the `GeometryUserData`

**Question** - in #4 its mentioned that an idea of the scene a ray is being fired upon would need to be stored in `RTCDRay`. I notice that `RTCDRayHit` already defines a `geomID`: 
```
struct RTCDHit : RTCHit {
  RTCDHit() {
    this->geomID = RTC_INVALID_GEOMETRY_ID;
    this->primID = RTC_INVALID_GEOMETRY_ID;
    this->Ng_x = 0.0;
    this->Ng_y = 0.0;
    this->Ng_z = 0.0;
  }
```
Is this `geomID` not that scene the ray is being fired upon? Should `RTCDRay` also additionally store a `geomID` or `sceneID`?

### DIFF
--- a/src/embree/ray_tracer.cpp
+++ b/src/embree/ray_tracer.cpp
@@ -62,9 +62,7 @@ TreeID EmbreeRayTracer::register_volume(const std::shared_ptr<MeshManager> mesh_
     // Check if this surface already has a cached surface geometry
     if (!surface_to_geometry_map_.count(surface)) 
     { // First visit: Create new RTCGeometry and GeometryUserData
-      auto registeredSurface = register_surface(mesh_manager, surface, volume_scene, storage_offset);
-      surface_geometry = registeredSurface.first;
-      surface_data = registeredSurface.second;
+      std::tie(surface_geometry, surface_data) = register_surface(mesh_manager, surface, volume_scene, storage_offset);
     } 
     else { // Second Visit: Recover existing RTCGeometry and GeometryUserData
       surface_geometry = surface_to_geometry_map_[surface];
@@ -76,10 +74,10 @@ TreeID EmbreeRayTracer::register_volume(const std::shared_ptr<MeshManager> mesh_
     }
 
     // Set the correct parent TreeID
-    auto parents = mesh_manager->get_parent_volumes(surface);
-    if (volume_id == parents.first) {
+    auto [forward_parent, reverse_parent] = mesh_manager->surface_senses(surface);
+    if (volume_id == forward_parent) {
       surface_data->forward_vol = tree;
-    } else if (volume_id == parents.second) {
+    } else if (volume_id == reverse_parent) {
       surface_data->reverse_vol = tree;
     } else {
       fatal_error("Volume {} is not a parent of surface {}", volume_id, surface);
--- a/src/embree/ray_tracer.cpp
+++ b/src/embree/ray_tracer.cpp
@@ -32,85 +32,82 @@ RTCScene EmbreeRayTracer::create_embree_scene() {
   rtcSetSceneBuildQuality(rtcscene, RTC_BUILD_QUALITY_HIGH);
   return rtcscene;
 }
-
-TreeID
-EmbreeRayTracer::register_volume(const std::shared_ptr<MeshManager> mesh_manager,
-                           MeshID volume_id)
+TreeID EmbreeRayTracer::register_volume(const std::shared_ptr<MeshManager> mesh_manager,
+                                        MeshID volume_id)
 {
   TreeID tree = next_tree_id();
   trees_.push_back(tree);
   auto volume_scene = this->create_embree_scene();
 
-  // allocate storage for this volume
-  auto volume_faces = mesh_manager->get_volume_faces(volume_id);
-  this->primitive_ref_storage_[volume_scene].resize(volume_faces.size());
-  auto& triangle_storage = this->primitive_ref_storage_[volume_scene];
-
   auto volume_surfaces = mesh_manager->get_volume_surfaces(volume_id);
-  int storage_offset {0};
+
+  // Get the number of prims that need to be added
+  size_t total_face_count = 0;
   for (auto& surface_id : volume_surfaces) {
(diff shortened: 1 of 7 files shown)


pshriwise:
> Question - in https://github.com/pshriwise/xdg/issues/4 its mentioned that an idea of the scene a ray is being fired upon would need to be stored in RTCDRay. I notice that RTCDRayHit already defines a geomID...

The `geomID` refers to a surface in the geometry model. The name mapping from geometry to Embree data space is as follows:

scene <--> volume
geometry <--> surface
primitive <--> triangle


This is probably worth adding to the developer's guide in the future. I'll make a note in the project board.

src/embree/ray_tracer.cpp
```diff
@@ -33,84 +33,82 @@ RTCScene EmbreeRayTracer::create_embree_scene() {
...
+        triangle_storage[storage_offset + i].primitive_id = surface_faces[i];
+      }
+
+      surface_geometry = rtcNewGeometry(device_, RTC_GEOMETRY_TYPE_USER);
+      rtcSetGeometryUserPrimitiveCount(surface_geometry, surf_face_count);
+      rtcAttachGeometry(volume_scene, surface_geometry);
+      surface_to_geometry_map_[surface] = surface_geometry;
+
+      surface_data = std::make_shared<GeometryUserData>();
+      surface_data->box_bump = bump;
```
pshriwise, line 79: One thing with this: 

Because this surface_data now applies to more than one volume, we need to use the larger of the two bump values for the volumes that are related to this surface to ensure the bounding boxes are dilated in a conservative manner based on how far a ray can travel within that volume.

I'll marked where we should do this this below.

src/embree/ray_tracer.cpp
```diff
@@ -33,84 +33,82 @@ RTCScene EmbreeRayTracer::create_embree_scene() {
...
+      rtcSetGeometryBoundsFunction(surface_geometry, (RTCBoundsFunction)&TriangleBoundsFunc, nullptr);
+      rtcSetGeometryIntersectFunction(surface_geometry, (RTCIntersectFunctionN)&TriangleIntersectionFunc);
+      rtcSetGeometryOccludedFunction(surface_geometry, (RTCOccludedFunctionN)&TriangleOcclusionFunc);
+      rtcCommitGeometry(surface_geometry);
+
+      storage_offset += surf_face_count;
+    } 
+    else { // Already exists: get geometry and user data
+      surface_geometry = surface_to_geometry_map_[surface];
+      surface_data = user_data_map_.at(surface_geometry);
```
pshriwise, line 96:
```suggestion
      surface_data = user_data_map_.at(surface_geometry);
      // set the box dilation value to the larger of the two box bump values for 
      // the volumes on either side of this surface
      surface_data->bump = std::max(surface_data->bump, bump);
```

gonuke (commented):
I have no comments here - I'll leave it to @pshriwise to merge

Waqar-ukaea:
> I like this as it simplifies the logic in register_volume. Would you be up for taking a crack at it?

I'm pretty happy with this one now @pshriwise. I've opted to make it a private method exclusive to `EmbreeRayTracer` but I could make it something that is available at the `RayTracer` interface level if I end up doing something similar in `GPRTRayTracer`.

src/embree/ray_tracer.cpp
```diff
@@ -33,88 +33,104 @@ RTCScene EmbreeRayTracer::create_embree_scene() {
...
+  int storage_offset = 0;
+
+  for (auto& surface : volume_surfaces) {
+    RTCGeometry surface_geometry;
+    std::shared_ptr<GeometryUserData> surface_data;
+
+    // Check if this surface already has a cached surface geometry
+    if (!surface_to_geometry_map_.count(surface)) 
+    { // First visit: Create new RTCGeometry and GeometryUserData
+      std::tie(surface_geometry, surface_data) = register_surface(mesh_manager, surface, volume_scene, storage_offset);
```
pshriwise, line 65: We'll want to set the value of `box_bump` here as well otherwise this volumes box dilation isn't taken into account.

Waqar-ukaea:
Whilst this does compile as it should, running the `ray_fire` tool with the largest model I have lying about (36093085 triangles as printed by h5minfo), I don't see a significant difference in the memory footprint between this PR and the main branch. They both seem to peak at about the same memory usage (using the python script you sent me for monitoring mem usage @pshriwise):

mem-usage

Waqar-ukaea:
Maybe by attaching the same `RTCGeometry` and `GeometryUserData` to a different `RTCScene`, embree is still internally duplicating the geometry? So despite only explicitly creating one set we don't actually see a measurable difference in memory usage? 

I did some reading on the embree API and came across this page on instancing - https://manpages.debian.org/experimental/libembree-dev/RTC_GEOMETRY_TYPE_INSTANCE.3.en.html which suggests that making use of `RTCGeometries` of `RTC_GEOMETRY_TYPE_INSTANCE` could be a way to reduce memory overhead because the instanced scene is "only stored one time". It seems like these are designed for situations where you have shared geometry. 

This sounds like it would be similar (not entirely sure if its the exact same system) to what I have been doing with GPRT where I create a BLAS and associated `gprt::instance` for each surface and my volume tree is a TLAS of those instances.

PR #58, opened 2025-02-10
Use minimal bump value

pshriwise:
#53 fixed a robustness problem, but introduced a performance problem by setting the box expansion value (a.k.a. `bump` variable) to 1.0. Not a great idea as this can cause significant overlap of bounding boxes, many false positive tests during BHV traversal, and performance degredation. 

This PR simply sets it to the intended value (the minimum of 1e-03 and the length of the bounding box diagonal for the problem at hand). This should resolve the issue reported in #4. Thanks again for catching this quickly @Waqar-ukaea!


pshriwise:
Looks like this took care of it. Care to confirm locally @Waqar-ukaea?

image

Waqar-ukaea:
Looks all good on my end!

Waqar-ukaea (approved):

PR #60, opened 2025-02-13
Allow raytracer setup for a single volume

pshriwise:
This is a minor refactor of the `XDG::prepare_raytracer` method that allows a single volume to be prepped. This is useful in tools like `ray_fire` where we really only need to build the BVH for a single volume. It should improve the responsiveness of such tools.


pshriwise:
Note: this also normalizes the input direction from the command line.

PR #62, opened 2025-02-13
Threaded (and general) performance-related changed

pshriwise:
This PR contains performance-critical updates for threaded (and non-threaded) scaling.

The change that affects threaded scalability is the update of the ray tracer accessor method to return a reference to the existing shared pointer instead of a shared pointer. Nothing is wrong with this in terms of behavior, but that object can't be operated on while a new shared pointer object is being created.

There's also an update to some of the XDG geometry to raytracer mapping from `std::map` to `std::unordered_map`. These structures are hit very often in ray queries and the O(1) lookup time from the hash table is needed.

### DIFF
--- a/include/xdg/xdg.h
+++ b/include/xdg/xdg.h
@@ -2,6 +2,7 @@
 #define _XDG_INTERFACE_H
 
 #include <memory>
+#include <unordered_map>
 
 #include "xdg/mesh_manager_interface.h"
 #include "xdg/ray_tracing_interface.h"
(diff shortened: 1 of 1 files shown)


pshriwise:
Some evidence of the changes here for a couple of geometries:

ATR_xdg_inactive_rates
ATR_xdg_active_rates

The rates after this update can be found in the [benchmarking dash](https://pshriwise.github.io/xdg-benchmarking/)

Issue #42, opened 2025-01-09
Expose orientation options for `ray_fire`


pshriwise:
There are situations in which we'd like to accept hits with orientations other than `HitOrientation::EXITING`.

For example, in the overlap checker it would be nice to be able to determine all surface intersections between volumes not just the on the far side of the volume wrt the ray being used to test for overlaps.

Waqar-ukaea:
Closed out by PR#63

PR #63, opened 2025-02-13
Expose ray orientation in ray_fire signature

Waqar-ukaea:
Resolving issue #42 by adding the ability to specify an orientation to the `ray_fire` call by passing an explicit `HitOrientation` enum.
```

  std::pair<double, MeshID> ray_fire(TreeID scene,
                                     const Position& origin,
                                     const Direction& direction,
                                     const double dist_limit = INFTY,
                                     std::vector<MeshID>* const exclude_primitives = nullptr,
                                     HitOrientation orientation = HitOrientation::EXITING);
```
Default value is `EXITING` as most queries will be exiting I assume?

Added some asserts to the `test_ray_fire.cpp` file for using `ray_fire` with a ray orientation specifying whether or not the ray is querying for a `ENTERING` or `EXITING` intersection. Diagram of the rays being tested for below:

image

@pshriwise Am I missing anything else or is this working as expected?

### DIFF
--- a/include/xdg/ray_tracing_interface.h
+++ b/include/xdg/ray_tracing_interface.h
@@ -41,8 +41,8 @@ public:
                                      const Position& origin,
                                      const Direction& direction,
                                      const double dist_limit = INFTY,
-                                     std::vector<MeshID>* const exclude_primitives = nullptr,
-                                     HitOrientation orientation = HitOrientation::EXITING);
+                                     HitOrientation orientation = HitOrientation::EXITING,
+                                     std::vector<MeshID>* const exclude_primitives = nullptr);
 
   void closest(TreeID scene,
                const Position& origin,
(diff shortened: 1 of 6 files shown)


Waqar-ukaea:
> One thing: even though it's a breaking change I'd probably prefer that this argument come before the `exclude_primities` argument so that a `nullptr` isn't required to set the orientation.

I think this makes a lot of sense. I'm not sure if the ray_fire signature will require many more optional parameters added later but as a general rule I will try to keep any optionals with nullptr defaults as the final parameter(s) passed to any of the core user facing API. 

The latest commit implements this change and should (hopefully) fix all the problems with switching the two parameters round - the tests are passing at the very least. 

@pshriwise I guess any applications you have making use of xdg downstream will be broken though. So I don't know if we want to hold off on merging this just yet?

pshriwise:
@Waqar-ukaea true true about the downstream issues. I think it's only going to require me changing the `xdg` branch in my fork of OpenMC (assuming no one else has randomly started using this). I'll keep an eye on CI here and make the changes as needed.

PR #66, opened 2025-02-24
Adding micro-benchmarks to XDG

Waqar-ukaea:
I think it would be a good idea if XDG also implemented micro-benchmarks alongside its unit tests for the various ray tracing operations so that we have a record of performance of these operations when code changes are made. 

This PR aims to add some micro-benchmarks for each of the ray tracing operations currently available in XDG. Catch2 supports benchmarking natively so I was thinking we could just go ahead and use the Catch2 macros rather than bringing in another library. Details of how to implement benchmarks with Catch2 can be found here - https://github.com/catchorg/Catch2/blob/devel/docs/benchmarks.md 

If we want more advanced benchmarking control something like https://github.com/google/benchmark may be more suitable though.

Some other considerations:
- We can just add some benchmarks to the existing test cases but it might make more sense to define some simple geometry/problems explicitly for benchmarking purposes.
- Each additional benchmark will ofc increase the time it takes for the CI to complete so will have to think about the length of these benchmarks and how many we actually need. 
- How these benchmarks actually interact with the CI will also need to be considered. Should large performance drops block merging for example.
- Perhaps these benchmarks could also be extended to some of the mesh operations too.
- Ctest by default won't display benchmark results unless called with `--verbose` as a benchmark won't count as failing.
- Benchmarking results could be logged to seperate file and used for analysis in within the workflow but could also be stored as github artifacts to allow some persistent history. (Probably makes sense for an artifact-based workflow to only be triggered on merges into main)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
         shell: bash
         run: |
           cd build
-          ctest -j4 --output-on-failure
+          ctest -j4 --output-on-failure --verbose
 
       - name: Check installed executables
         shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
         shell: bash
         run: |
           cd build
-          ctest -j4 --output-on-failure --verbose
+          ctest -j1 --output-on-failure --verbose
 
       - name: Check installed executables
         shell: bash
--- a/tests/test_point_in_volume.cpp
+++ b/tests/test_point_in_volume.cpp
@@ -4,6 +4,8 @@
 // xdg includes
 #include "xdg/mesh_manager_interface.h"
 #include "xdg/ray_tracing_interface.h"
+#include <catch2/benchmark/catch_benchmark.hpp>
+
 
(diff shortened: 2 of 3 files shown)


Waqar-ukaea:
A minor thing but since we are calling the tests in parallel with 4 processes the outputs aren't kept together. A fix for this could just be to define the benchmarks as a separate test case entirely which should (hopefully) ensure they do stay together. Otherwise we could just switch to testing in serial.

pshriwise:
This is nice @Waqar-ukaea! Thanks for exploring this. 

One thought: CI machines can change and I don't think we're guaranteed a specific machine for a given run, making performance comparisons tricky. I really like the idea of building these benchmarks into the test suite, but we may want to limit comparisons to runs on a consistent machine. I'm looking into how we might connect to one, but it could take some time to ensure we're doing it securely (it's behind a firewall).

Waqar-ukaea:
Sounds good @pshriwise. In that case I can carry on with what I had in mind for this PR and we can keep it unmerged until we figure out what we're doing with the CI machine.

Waqar-ukaea:
@pshriwise happy for me to mark this PR as closed since we never ended up exploring further?

PR #88, opened 2025-03-21
Method nomenclature

pshriwise:
This PR switches some of the method naming so that methods with `*_faces` refer to 2D elements and `*_elements` refers to volumetric, 3D elements.

This is primarily to disambiguate the methods that collect all of the volume triangles from methods that I'll introduce to collect all of the volumetric elements when particles are walking element-to-element.

### DIFF
--- a/include/xdg/libmesh/mesh_manager.h
+++ b/include/xdg/libmesh/mesh_manager.h
@@ -292,7 +292,7 @@ public:
 
   // sideset face mapping, stores the element and the side number
   // for each face in the mesh that lies on a boundary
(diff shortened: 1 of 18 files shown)


gonuke:
I wasn't watching when this PR when through, so it's probably too late given subsequent development... but.... `facets` might have been more clear than `faces` IMO.  I thought I'd leave it here in case there was any incentive to change :)

pshriwise:
Not too late! Your input is appreciated. But given that we're hoping to look into support for second order elements I think faces is a little more generic in a way that captures that scope.

PR #97, opened 2025-04-08
UML Diagram Updates

pshriwise:
Indent the `<div>` line so it's recognized as part of the raw HTML block in the RST file. 

I also slid this below the initial text. Hope that's okay. I think the language above provides some helpful context for the design in the UML.


Waqar-ukaea:
> I also slid this below the initial text. 

Happy with this, I think it makes sense to have the diagram afterwards so the reader has the context from the main bulk of the text.

PR #100, opened 2025-04-15
Expose methods to return MOAB vertices and connectivity

Waqar-ukaea:
This PR adds methods to the mesh manager interface to return the coordinates of vertices in a MOAB mesh as well as the connectivity between those vertices. The motivation for publicly exposing this is to make it easier to use the `MeshManager` API to transfer element vertices/indices to a GPRT buffer for building acceleration structures on device. 

Two methods have been added:
- `MeshManager::get_surface_vertices(MeshID surface)` returns a flattened std::vector<double> of xyz coords of all the vertices in a surface.
- `MeshManager::get_surface_connectivity(MeshID surface)` returns a flattened std::vector<int> of indices for each vertex in a surface.

I've also added a couple tests to `test_moab.cpp` to ensure that these methods are returning the correct the number of vertices/indices. And added stubs for the `LibMeshManager`.

@pshriwise please let me know if you feel like this shouldn't be exposed in the `MeshManager` or if it makes more sense for the methods to return non-flattened vectors, etc...

### DIFF
--- a/include/xdg/mesh_manager_interface.h
+++ b/include/xdg/mesh_manager_interface.h
@@ -46,6 +46,11 @@ public:
 
   virtual std::array<Vertex, 3> face_vertices(MeshID element) const = 0;
 
+  virtual std::vector<int> get_surface_connectivity(MeshID surface) const = 0;
+
+  virtual std::vector<double> get_surface_vertices(MeshID surface) const = 0;
+
+
   BoundingBox element_bounding_box(MeshID element) const;
 
   BoundingBox face_bounding_box(MeshID element) const;
--- a/include/xdg/moab/mesh_manager.h
+++ b/include/xdg/moab/mesh_manager.h
@@ -64,6 +64,10 @@ public:
 
   std::array<Vertex, 3> face_vertices(MeshID element) const override;
 
+  std::vector<int> get_surface_connectivity(MeshID surface) const;
+
+  std::vector<double> get_surface_vertices(MeshID surface) const;
+
   // Topology
   std::pair<MeshID, MeshID> surface_senses(MeshID surface) const override;
 
--- a/include/xdg/moab/mesh_manager.h
+++ b/include/xdg/moab/mesh_manager.h
@@ -64,9 +64,9 @@ public:
 
   std::array<Vertex, 3> face_vertices(MeshID element) const override;
 
-  std::vector<int> get_surface_connectivity(MeshID surface) const;
+  std::vector<int> get_surface_connectivity(MeshID surface) const override;
 
-  std::vector<double> get_surface_vertices(MeshID surface) const;
+  std::vector<double> get_surface_vertices(MeshID surface) const override;
 
   // Topology
   std::pair<MeshID, MeshID> surface_senses(MeshID surface) const override;
--- a/src/moab/mesh_manager.cpp
+++ b/src/moab/mesh_manager.cpp
@@ -269,6 +269,35 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
   return surface_ids;
 }
 
+std::vector<int> 
+MOABMeshManager::get_surface_connectivity(MeshID surface) const
+{
+  moab::EntityHandle surf_handle = this->surface_id_map_.at(surface);
+  auto faces = _surface_faces(surface);
+
+  moab::Range conn;
+  this->moab_interface()->get_connectivity(faces, conn);
+
+  std::vector<int> connectivity;
+  connectivity.insert(connectivity.end(), conn.begin(), conn.end());
+
+  return connectivity;
+}
+
+std::vector<double> 
+MOABMeshManager::get_surface_vertices(MeshID surface) const
+{
+  auto conn = get_surface_connectivity(surface);
+  std::vector<moab::EntityHandle> verts;
+
+  verts.insert(verts.end(), conn.begin(), conn.end());
+  std::vector<double> coords(verts.size() * 3);
+
+  this->moab_interface()->get_coords(verts.data(), verts.size(), coords.data());
+
+  return coords;
+}
+
 void
 MOABMeshManager::parse_metadata()
 {
--- a/tests/test_moab.cpp
+++ b/tests/test_moab.cpp
@@ -111,6 +111,36 @@ TEST_CASE("Test Ray Fire MOAB")
 
 }
 
+TEST_CASE("MOAB Vertices")
+{
+  std::shared_ptr<XDG> xdg = XDG::create(MeshLibrary::MOAB);
+  REQUIRE(xdg->mesh_manager()->mesh_library() == MeshLibrary::MOAB);
+  const auto& mesh_manager = xdg->mesh_manager();
+  mesh_manager->load_file("cube.h5m");
+  mesh_manager->init();
+
+  int vertex_count = 0;
+
+  for (const auto surface: mesh_manager->surfaces()) {
+    auto vertices = mesh_manager->get_surface_vertices(surface);
+    REQUIRE(vertices.size() / 3 == 138); // Each surface should contain exactly 138 nodes
+  }
+}
+
+TEST_CASE("MOAB Connectivity")
+{
+  std::shared_ptr<XDG> xdg = XDG::create(MeshLibrary::MOAB);
+  REQUIRE(xdg->mesh_manager()->mesh_library() == MeshLibrary::MOAB);
+  const auto& mesh_manager = xdg->mesh_manager();
+  mesh_manager->load_file("cube.h5m");
+  mesh_manager->init();
+
+  for (const auto surface: mesh_manager->surfaces()) {
+    auto connectivity = mesh_manager->get_surface_connectivity(surface);
+    REQUIRE(connectivity.size() == 138); // Each surface should contain exactly 138 nodes
+  }
+}
+
 TEST_CASE("TEST XDG Factory Method")
 {
 
--- a/include/xdg/libmesh/mesh_manager.h
+++ b/include/xdg/libmesh/mesh_manager.h
@@ -99,6 +99,16 @@ public:
 
   std::vector<MeshID> get_volume_surfaces(MeshID volume) const override;
 
+  std::vector<int> get_surface_connectivity(MeshID surface) const override
+  {
+    fatal_error("LibMeshManager::get_surface_connectivity() not implemented yet");
+  }
+
+  std::vector<double> get_surface_vertices(MeshID surface) const override
+  {
+    fatal_error("LibMeshManager::get_surface_vertices() not implemented yet");
+  }
+
   MeshID create_volume() override;
 
   void add_surface_to_volume(MeshID volume, MeshID surface, Sense sense, bool overwrite=false) override;
--- a/tests/mesh_mock.h
+++ b/tests/mesh_mock.h
@@ -75,6 +75,32 @@ public:
     return {vertices[0], vertices[1], vertices[2]};
   }
 
+  std::vector<int> get_surface_connectivity(MeshID surface) const override
+  {
+    std::vector<int> flat_connectivity;
+    flat_connectivity.reserve(triangle_connectivity.size() * 3);
+    
+    for (const auto& tri : triangle_connectivity) {
+      flat_connectivity.insert(flat_connectivity.end(), tri.begin(), tri.end());
+    }
+
+    return flat_connectivity;
+  }
+
+  std::vector<double> get_surface_vertices(MeshID surface) const override
+  {
+    std::vector<double> flat_vertices;
+    flat_vertices.reserve(vertices.size() * 3);
+    
+    for (const auto& vertex : vertices) {
+      flat_vertices.push_back(vertex.x);
+      flat_vertices.push_back(vertex.y);
+      flat_vertices.push_back(vertex.z);
+    }
+
+    return flat_vertices;
+  }
+
   // Topology
   virtual std::pair<MeshID, MeshID> surface_senses(MeshID surface) const override {
     return {0, ID_NONE};


include/xdg/mesh_manager_interface.h
```diff
@@ -46,6 +46,11 @@ class MeshManager {
 
   virtual std::array<Vertex, 3> face_vertices(MeshID element) const = 0;
 
+  virtual std::vector<int> get_surface_connectivity(MeshID surface) const = 0;
+
+  virtual std::vector<double> get_surface_vertices(MeshID surface) const = 0;
```
pshriwise, line 51: Does it make sense to return a vector of `xdg:Vertex` here?

include/xdg/moab/mesh_manager.h
```diff
@@ -64,6 +64,10 @@ class MOABMeshManager : public MeshManager {
 
   std::array<Vertex, 3> face_vertices(MeshID element) const override;
 
+  std::vector<int> get_surface_connectivity(MeshID surface) const override;
```
pshriwise, line 67: I'm trying to be forward-thinking about times when we'll be supporting other elements.

Depending on the element type, we'll need to know how many connectivity values each element uses, so we may want to return that as well. Or, we could add a method to retrieve the surface element type? (hardcoded to triangles for now)

pshriwise (commented):
A couple of design thoughts to discuss, but nothing major.

We'll need to think about how to produce connectivity for libMesh surfaces at some point.

include/xdg/mesh_manager_interface.h
Waqar-ukaea: Yep, totally happy with that. I needed flattened arrays to pass to GPRT buffers but I think returning a vector of `Vertex` and flattening the array afterwards is fine too.

Waqar-ukaea:
> We'll need to think about how to produce connectivity for libMesh surfaces at some point.

Does libmesh track any kind of connectivity between elements? Maybe called something else and not connectivity in libmesh syntax?

include/xdg/moab/mesh_manager.h
Waqar-ukaea: I think having a method to retrieve the surface element type would be useful. 

More so just for curiosity (not sure how it would be useful) but would there ever be a scenario where we read a mes into XDG composed of multiple element types?

Waqar-ukaea:
I've rewritten these methods and also written a new method that returns both vertices and connectivity since I was struggling to retain the correct local connectivity for a surface when using the individual methods. Not sure if I want to keep the individual methods or just stick with the combined `MeshManager::get_surface_mesh` method.

I wanted to write a test for `get_surface_mesh` but I wasn't sure how to go about doing so without brute force testing each element so that's what I have done so far. `MeshManager` reads in a fairly small mesh `overlap-edge.h5m` and does a comparison between a flattened vector of connectivity against the expected and the same for vertices in xyz. 

I actually already merged these changes into my GPRT branch and have successfully used them to ray trace against a MOAB mesh in single precision so this is honestly probably ready to be pushed into main - see https://github.com/pshriwise/xdg/pull/94#issuecomment-2841553742 for updates on that. @pshriwise just let me know if you are happy with the changes made and i'll merge.

tests/test_moab.cpp
```diff
@@ -111,6 +111,128 @@ TEST_CASE("Test Ray Fire MOAB")
 
 }
 
+// TEST_CASE("MOAB Vertices")
```
pshriwise, line 114: Looks like this commented code can be removed

tests/test_moab.cpp
```diff
@@ -111,6 +111,128 @@ TEST_CASE("Test Ray Fire MOAB")
...
+  std::shared_ptr<XDG> xdg = XDG::create(MeshLibrary::MOAB);
+  REQUIRE(xdg->mesh_manager()->mesh_library() == MeshLibrary::MOAB);
+  const auto& mesh_manager = xdg->mesh_manager();
+  mesh_manager->load_file("overlap-edge.h5m");
+  mesh_manager->init();
+
+  float fpTol = 1e-5;
+
+  // Define the expected connectivity and vertices for each surface
+  std::vector<std::vector<int>> expected_connectivity = {
```
pshriwise, line 152: As discussed last week, let's rely on the `MeshMock` model for these explicit checks.

pshriwise (changes requested):
Just a couple of small things here. The switch to `MeshMock` isn't such a small lift but I appreciate you taking the time to make that change. Thanks @Waqar-ukaea!

include/xdg/constants.h
```diff
@@ -103,6 +103,17 @@ enum class RayFireType { VOLUME, POINT_CONTAINMENT, ACCUMULATE_HITS, FIND_VOLUME
 //
 enum class HitOrientation { ANY, EXITING, ENTERING };
 
+// Enumerator for different element types (maybe we want more here?)
+enum class SurfaceElementType {
+  TRI = 0,
+  QUAD = 1,
+};
+
+enum class VolumeElementType {
```
gonuke, line 112: Maybe also a wedge?

This seems like a place to think ahead a little so that the enum ordering makes sense as new things are added

src/moab/mesh_manager.cpp
```diff
@@ -269,6 +270,105 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
...
+MOABMeshManager::get_surface_connectivity(MeshID surface) const
+{ 
+  std::vector<int> connectivity;
+  auto faces = get_surface_faces(surface);
+  int nodes;
+  for (auto face : faces) { 
+    moab::EntityHandle element_handle;
+    std::vector<EntityHandle> conn;
+    this->moab_interface()->handle_from_id(moab::MBTRI, face, element_handle);
+    this->moab_interface()->get_connectivity(&element_handle, 1, conn);
```
gonuke, line 283: MOAB can do this in a single call if you have a vector of entity handles... I also note that most of what `get_surface_faces()` does is to convert entity handles to IDs, and then you need to convert them back here... I guess this may be motivated by other design choices and being mesh-agnostic, but definitely burdensome for the MOAB case

src/moab/mesh_manager.cpp
```diff
@@ -269,6 +270,105 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
...
+  return connectivity;
+}
+
+std::vector<Vertex> 
+MOABMeshManager::get_surface_vertices(MeshID surface) const
+{
+  auto conn = get_surface_connectivity(surface);
+  std::vector<moab::EntityHandle> verts;
+
+  verts.insert(verts.end(), conn.begin(), conn.end());
```
gonuke, line 297: Unless I'm missing something, this will have many/most vertices represented multiple times - is that the desired outcome?  `get_surface_connectivity()` appears to get a list of the vertex EntityHandles for each triangle.  In most surfaces, each vertex will appear in many triangles, so that this connectivity list will include each vertex multiple times.

Perhaps you mean for `verts` to be a `std::set`? But there are probably more efficient ways to get the vertices from a surface (at least in MOAB)

src/moab/mesh_manager.cpp
```diff
@@ -269,6 +270,105 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
...
+    this->moab_interface()->get_connectivity(&element_handle, 1, conn);
+    connectivity.push_back(conn[0]);
+    connectivity.push_back(conn[1]);
+    connectivity.push_back(conn[2]);
+    verts.insert(verts.end(), conn.begin(), conn.end());
+  }
+
+  // Remove duplicate vertices
+  std::sort(verts.begin(), verts.end());
+  verts.erase(std::unique(verts.begin(), verts.end()), verts.end());
```
gonuke, line 330: You could simply declare `verts` as a `std::set` to avoid removing duplicates.

src/moab/mesh_manager.cpp
```diff
@@ -269,6 +270,105 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
...
+  std::vector<Vertex> vertices(verts.size());
+  for (int i = 0; i < verts.size(); i++) {
+    vertices[i] = Vertex(coords[3*i], coords[3*i+1], coords[3*i+2]);
+  }
+
+  return vertices; 
+}
+
+std::pair<std::vector<Vertex>, std::vector<int>> 
+MOABMeshManager::get_surface_mesh(MeshID surface) const
```
gonuke, line 311: This method could/should probably reuse the other methods that you already have to avoid duplicate code

src/moab/mesh_manager.cpp
```diff
@@ -269,6 +270,105 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
...
+
+  // Map vertices to their coordinates
+  std::vector<Vertex> vertices(verts.size());
+  for (size_t i = 0; i < verts.size(); i++) {
+    vertices[i] = Vertex(coords[3 * i], coords[3 * i + 1], coords[3 * i + 2]);
+  }
+
+  // Remap connectivity indices to local vertex indices
+  std::map<moab::EntityHandle, int> handle_to_index;
+  for (size_t i = 0; i < verts.size(); i++) {
```
gonuke, line 344: Combine this loop with the previous one?

src/moab/mesh_manager.cpp
gonuke: I see below that you do remove the duplicates in that method.  Probably should do the same here.  Since I expect there to be more duplicates than non-duplicates, it may be valuable to avoid adding the duplicates in the first place.

gonuke (changes requested):
I may have missed some important design discussions, but I think there are better ways to get the info from MOAB.

Most comments here are about interacting with the MOAB interface a little more cleanly.

include/xdg/constants.h
pshriwise: Is there any complication in adding element as they become supported? Even re-ordering shouldn't be an issue I don't think.

src/moab/mesh_manager.cpp
pshriwise: Hrmm yes. As you thought, `get_surface_faces` is a generic interface-level function using a placeholder ID type (`int32_t` for now) in place of the library-specific identifiers.

I'm not too worried as this method is really only called during setup and isn't performance-critical, but given that we do have an internal data member that stores the exact information we're looking for (`_surface_faces`) we can probably use that.

pshriwise:
> Does libmesh track any kind of connectivity between elements? Maybe called something else and not connectivity in libmesh syntax? Or would we need to add this ourselves?

Sorry I realized I never replied to this. libMesh does contain connectivity information, we'll just need to collect it element-by-element (as far as I understand it). libMesh doesn't have the batch collection methods that MOAB does.

Waqar-ukaea:
Hi @gonuke @pshriwise, thanks for the suggestions I've gone back through the code where I use these methods downstream and realised that I don't necessarily need the `get_surface_connectivity()` or `get_surface_vertices()` methods. The problem with both of these functions is that they return global (to the MOAB instance) ids for connectivity/indices. But for my usecase, in providing GPRT with vertices and triangle indices, I need local (to the surface) indices, requiring me to map from the global connectivity - hence why I also wrote `get_surface_mesh()` which returns the local indices along with coords of those vertices in the correct order. 

I can't really see a way to decouple this logic into two separate functions as i originally intended since I need the vertices to get the local indices. I have opted to remove the `get_surface_connectivity()` entirely and have kept `get_surface_vertices()` to return vertices without indices if that is ever required.

I have also rewritten parts of those methods to (hopefully) make use of the MOAB API in a cleaner fashion, making use of `_surface_vertices() and `moab_interface()->get_adjacencies()` to return vertices without duplicates.

tests/test_moab.cpp
Waqar-ukaea: @pshriwise should this test still live in the test_moab.cpp once switched out for `MeshMock`?

pshriwise: It should probably move into `test_mesh_internal`. I don't think it hurts to keep this MOAB-specific test, btw. 

My main concern here is that our test coverage of the interfaces isn't highly dependent on which optional dependencies are enabled.

pshriwise:
> Hi @gonuke @pshriwise, thanks for the suggestions I've gone back through the code where I use these methods downstream and realised that I don't necessarily need the `get_surface_connectivity()` or `get_surface_vertices()` methods. The problem with both of these functions is that they return global (to the MOAB instance) ids for connectivity/indices. But for my usecase, in providing GPRT with vertices and triangle indices, I need local (to the surface) indices, requiring me to map from the global connectivity - hence why I also wrote `get_surface_mesh()` which returns the local indices along with coords of those vertices in the correct order.
> 
> I can't really see a way to decouple this logic into two separate functions as i originally intended since I need the vertices to get the local indices. I have opted to remove the `get_surface_connectivity()` entirely and have kept `get_surface_vertices()` to return vertices without indices if that is ever required.
> 
> I have also rewritten parts of those methods to (hopefully) make use of the MOAB API in a cleaner fashion, making use of `_surface_vertices() and `moab_interface()->get_adjacencies()` to return vertices without duplicates.

Thank you @Waqar-ukaea! I'll have another look shortly, but what you've said makes sense. Down the road, I think accessing this info from the `MBDirectAccess` class will be more efficient as the triangle/element connectivity maps to the global ID space of the triangles naturally there, but I really don't want to lose focus of the main goal for this PR -- establishing ray tracing with GPRT.

src/moab/mesh_manager.cpp
```diff
@@ -2,6 +2,7 @@
 #include <map>
 #include <memory>
 #include <string>
+#include <set>
```
gonuke, line 5: Looks like you didn't use this after all :) so it's probably not needed here.

src/moab/mesh_manager.cpp
```diff
@@ -269,6 +270,78 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
...
+
+  // Retrieve vertex coordinates
+  std::vector<double> coords(verts.size() * 3);
+  this->moab_interface()->get_coords(verts, coords.data());
+
+  // Map vertices to their coordinates
+  std::vector<Vertex> vertices(verts.size());
+  for (size_t i = 0; i < verts.size(); i++) {
+    vertices[i] = Vertex(coords[3 * i], coords[3 * i + 1], coords[3 * i + 2]);
+  }
```
gonuke, line 323: If this happened before the connectivity, it would be identical to `get_surface_vertices`. Perhaps these could be modularized with that in mind?

gonuke: Or make the code to go from vertex handles to vertex coordinates a small internal function that you call in both places...

src/moab/mesh_manager.cpp
```diff
@@ -269,6 +270,78 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
...
+  std::vector<double> coords(verts.size() * 3);
+  this->moab_interface()->get_coords(verts, coords.data());
+  std::vector<Vertex> vertices(verts.size());
+
+  for (int i = 0; i < verts.size(); i++) {
+    vertices[i] = Vertex(coords[3*i], coords[3*i+1], coords[3*i+2]);
+  }
+
+  return vertices; 
+}
```
gonuke, line 288: Something like this will allow you to reuse it below...
```suggestion
   return _get_coords(verts);
}

std::vector<Vertex> MOABMeshManager::_get_coords(moab::Range& verts) {

  std::vector<double> coords(verts.size() * 3);
  this->moab_interface()->get_coords(verts, coords.data());
  std::vector<Vertex> vertices(verts.size());

  for (int i = 0; i < verts.size(); i++) {
    vertices[i] = Vertex(coords[3*i], coords[3*i+1], coords[3*i+2]);
  }

  return vertices; 
}
```

src/moab/mesh_manager.cpp
gonuke: with above change:
```suggestion
  vertices = _get_coords(verts);
```

gonuke (changes requested):
These are good changes @Waqar-ukaea - I have one small  suggestion for modularity.

src/moab/mesh_manager.cpp
Waqar-ukaea: I think this is a nice way to do things, so I'll go with this moving forwards.

pshriwise (approved):
I'm happy moving forward with the changes here, knowing that there will likely be subsequent changes to these methods as we navigate the execution of this work on GPU.

pshriwise:
@gonuke if you're happy with this after another look feel free to approve and merge.

tests/test_moab.cpp
```diff
@@ -111,6 +111,117 @@ TEST_CASE("Test Ray Fire MOAB")
...
+      {3, 0, 4, 5, 1, 3, 5, 3, 4, 5, 4, 2},                   /* Surface 5 */
+      {4, 5, 6, 0, 5, 3, 7, 2, 4, 4, 3, 5, 7, 4, 6, 7, 6, 1}, /* Surface 6 */
+      {6, 4, 5, 3, 5, 4, 1, 5, 3, 7, 0, 4, 7, 4, 6, 7, 6, 2}, /* Surface 7 */
+      {3, 6, 4, 1, 5, 3, 7, 2, 4, 3, 5, 6, 7, 4, 6, 7, 6, 0}  /* Surface 8 */
+  };
+
+  std::vector<std::vector<Vertex>> expected_vertices = {
+      // Surface 1
+      {
+          {0.5, -0.866025, -1.5}, {-1, 1.22465e-16, -1.5}, {0.5, 0.866025, -1.5},
```
gonuke, line 153: If you are testing within some FP tolerance below, can/should this 1.22465e-16 be 0? Just curious...

gonuke (approved):
I have a couple of non-blocking queries here in tests.  I'll wait to merge in case there are quick responses - feel free to let me know that you'd prefer to leave it as is and I'll merge.

tests/test_moab.cpp
```diff
@@ -111,6 +111,117 @@ TEST_CASE("Test Ray Fire MOAB")
...
+    for (size_t i = 0; i < connectivity.size(); ++i) {
+      REQUIRE(connectivity[i] == expected_connectivity[surface_index][i]);
+    }
+
+    // Test vertices
+    REQUIRE(vertices.size() == expected_vertices[surface_index].size());
+    for (size_t i = 0; i < vertices.size(); ++i) {
+      REQUIRE_THAT(vertices[i].x, Catch::Matchers::WithinAbs(expected_vertices[surface_index][i].x, fpTol));
+      REQUIRE_THAT(vertices[i].y, Catch::Matchers::WithinAbs(expected_vertices[surface_index][i].y, fpTol));
+      REQUIRE_THAT(vertices[i].z, Catch::Matchers::WithinAbs(expected_vertices[surface_index][i].z, fpTol));
```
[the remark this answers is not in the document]
Waqar-ukaea: I just checked our tests in XDG and I didn't see any of them using  `Catch::Approx` so I guess you were referring to other repos?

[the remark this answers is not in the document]
gonuke: Oh - I just approved/merged a PR from @pshriwise and assumed it was ubiquitous... I guess someone can decide in the future whether to streamline these

PR #102, opened 2025-04-22
Initial volumetric mesh tracking on tets

pshriwise:
This contains an initial implementation of volumetric mesh tracking on linear tets. Embree's BVH is currently used for point location purposes. More information on this can be found in #30.

TODOs:

  - [x] Implement a MOAB direct access lookup for tetrahedron coordinates
  - [x] Add direct access lookup for element adjacencies (if needed)
  - [x] Add an interface-level method to determine given element's domain ID
    - [x] + tests
  - [x] Separate volumetric point location into its own class (so different implementations can be used)
  - [ ] Evaluate performance of equivalent MOAB and libMesh models


pshriwise:
This PR has become too large to reasonably ask someone to review so I'm going to break it up into smaller PRs. This is not an exhaustive list but some PRs will include. 

- [ ] Add linalg as a new vendored dependency
- [ ] Add OpenMP and wrapped interface to support setting threads for libMesh (not entirely happy with this situation)
- [ ] Addition of timers for mini-apps
- [ ] new `Vec3da` methods, formatter, and tests
- [ ] Faster `next_sidepair_id`
- [ ] `BoundingBox` `fmt::formatter`
- [ ] MOAB element adjacencies via direct access
- [ ] Addition of tet containment method and tests
- [ ] Addition of Embree intersector functions
- [ ] Addition of `RTCElementRay`
- [ ] Addition of Embree trees for tet elements ( and `VolumeElementsUserData`)
- [ ] Reporting precompiler definitions (maybe better replaced with a clever logger)
- [ ] Addition of `next_element` methods for mesh managers
- [ ] Addition of `MeshManagerInterface::walk_elements`
- [ ] Exposure of `XDG::segements` and `XDG::next_element`

I'll track them here. This list may change as I add or remove them.

In the meantime I can use this branch to start design of related classes in OpenMC.

PR #104, opened 2025-04-28
Added [[Noreturn]] attribute on XDG fatal errors

Waqar-ukaea:
Super minor PR that adds the [[noreturn]] attribute to the `fatal_error()` function defined in `error.h`. 

Since `fatal_error()` always results in exiting the application the control logic will never return to the caller - `[[noreturn]]` explicitly specifies this to the compiler, which should suppress the compiler warning: `warning: no return statement in function returning non-void [-Wreturn-type]`.


pshriwise (approved):
Looks good! Thanks!

PR #107, opened 2025-05-12
Refactor xdg create

Waqar-ukaea:
Refactored the `xdg::create()` method to make use of a set of lambda functions for extra encapsulation in setting the back-end libraries XDG is instantiated with. Should hopefully allow for a cleaner more extensible structure as more back-end libraries are added.


Waqar-ukaea:
Not sure why the tests are failing here and not in the GPRT branch but once I get them fixed here I will go ahead and merge this into main.

PR #109, opened 2025-05-20
Update progress bar implementation

pshriwise:
Using a third party progress bar that looks a bit nicer than our in-house one borrowed from OpenMC.


src/overlap_check/overlap.cpp
```diff
@@ -127,13 +139,24 @@ void check_instance_for_overlaps(std::shared_ptr<XDG> xdg,
...
   }
 
-  std::cout << "Checking for overlapped regions along element edges..." << std::endl;
+  // Number of rays cast along edges = number_of_elements * (number_of_edges * 2) * (number_of_vols - parent_vols)
+  int totalEdgeRays = totalElements*(3)*(allVols.size()-2);
+
+  std::cout << fmt::format("Checking for overlapped regions along {} element edges...", totalEdgeRays) << std::endl;
+
+  using namespace indicators;
+  BlockProgressBar edge_bar{
```
gonuke, line 148: You're one progress bar short of me suggesting a method that initializes these with your standard configuration and only passes a couple of variables... :)

tools/CMakeLists.txt
```diff
@@ -12,6 +12,6 @@ overlap_check
 foreach(tool ${TOOL_NAMES})
     string(REPLACE "_" "-" tool_exec ${tool})
     add_executable(${tool_exec} ${tool}.cpp)
-    target_link_libraries(${tool_exec} xdg argparse)
+    target_link_libraries(${tool_exec} xdg argparse indicators::indicators)
```
gonuke, line 15: this syntax is new to me... what does the `::` reference?  This is also a header-only progress bar isn't it (or at least that's what I took away from it's GH README)? so even more confused about what you link to.

pshriwise: It's like a namespace in C++, but for CMake targets: `project::target`.

In this case they happen to be the same name.

src/overlap_check/overlap.cpp
pshriwise: It isn't so clear here, but this will be used in other tools/mini-apps in `tets-initial`.

Waqar-ukaea (approved):
Looks good to me! Happy to merge if you are @pshriwise.

pshriwise:
Go for it!

src/overlap_check/overlap.cpp
```diff
@@ -127,13 +127,16 @@ void check_instance_for_overlaps(std::shared_ptr<XDG> xdg,
     return;
   }
 
-  std::cout << "Checking for overlapped regions along element edges..." << std::endl;
+  // Number of rays cast along edges = number_of_elements * (number_of_edges * 2) * (number_of_vols - parent_vols)
+  int totalEdgeRays = totalElements*(3)*(allVols.size()-2);
+
+  std::cout << fmt::format("Checking for overlapped regions along {} element edges...", totalEdgeRays) << std::endl;
+
+  using namespace indicators;
```
gonuke, line 135: You don't need this namespace here anymore, right?

gonuke (changes requested):
Looks like one leftover namespace after refactor

src/overlap_check/overlap.cpp
pshriwise: Correct. Thanks! Taken care of.

gonuke (approved):
Looks good - will merge after tests

pshriwise:
Added to the installed configuration to ensure the `indicators` targets are pulled in by downstream projects.

Downstream doesn't really have need of this library, so in the future it would be good to separate this from `libxdg`, but in the interest of progress and focus on other capabilities I'll create an issue and table that task for now. #110

gonuke:
I don't appear to have permission to merge :)

Issue #111, opened 2025-06-02
Add exception handling to metadata parsing for cases where no metadata found


Waqar-ukaea:
Right now `MeshManager::parse_metadata()` leads to a segfault because an `out_of_range` exception is thrown when no metadata is found in the geometry. 

I think we should probably be calling `parse_metadata()` by default every time the geometry is initialised as good practice - it would be nice to handle this exception so execution can continue when there is no metadata found. 

Adding some tests here will also be useful since we will probably want to build out the metadata parsing capabilities within XDG in the future.

Backtrace leading to segfault:
![Image](https://github.com/user-attachments/assets/57b87b55-afb4-478c-9498-04e5db0babc9)

PR #112, opened 2025-06-03
Parse metadata exception handling

Waqar-ukaea:
PR for #111

### DIFF
--- a/src/mesh_manager_interface.cpp
+++ b/src/mesh_manager_interface.cpp
@@ -82,9 +82,14 @@ MeshManager::get_volume_property(MeshID volume, PropertyType type) const
 Property
 MeshManager::get_surface_property(MeshID surface, PropertyType type) const
 {
-  if (surface_metadata_.count({surface, type}) == 0)
-    return {PropertyType::BOUNDARY_CONDITION, "transmission"};
-  return surface_metadata_.at({surface, type});
+  try {
+    if (surface_metadata_.count({surface, type}) == 0)
+      return {PropertyType::BOUNDARY_CONDITION, "transmission"};
+    return surface_metadata_.at({surface, type});
+  }
+  catch (const std::out_of_range e) {
+    return VOID_MATERIAL;
+  }
 }
 
 MeshID MeshManager::next_volume(MeshID current_volume, MeshID surface) const
--- a/src/mesh_manager_interface.cpp
+++ b/src/mesh_manager_interface.cpp
@@ -71,7 +71,12 @@ MeshManager::surface_has_property(MeshID surface, PropertyType type) const
 Property
 MeshManager::get_volume_property(MeshID volume, PropertyType type) const
 {
-  return volume_metadata_.at({volume, type});
+  try { 
+    return volume_metadata_.at({volume, type});
+  }
+  catch (const std::out_of_range e) {  
+    return VOID_MATERIAL;
+  }
 }
 
 Property


src/mesh_manager_interface.cpp
```diff
@@ -71,15 +71,25 @@ MeshManager::surface_has_property(MeshID surface, PropertyType type) const
...
-  if (surface_metadata_.count({surface, type}) == 0)
-    return {PropertyType::BOUNDARY_CONDITION, "transmission"};
-  return surface_metadata_.at({surface, type});
+  try {
+    if (surface_metadata_.count({surface, type}) == 0)
+      return {PropertyType::BOUNDARY_CONDITION, "transmission"};
+    return surface_metadata_.at({surface, type});
+  }
+  catch (const std::out_of_range e) {
+    return VOID_MATERIAL;
```
gonuke, line 91: This already suggests that the transmission boundary condition should be the default. A surface property is never a material. 

Although I wonder if this is a good choice for default if/when other properties get applied to surfaces?

src/mesh_manager_interface.cpp
```diff
@@ -71,15 +71,25 @@ MeshManager::surface_has_property(MeshID surface, PropertyType type) const
...
 {
-  if (surface_metadata_.count({surface, type}) == 0)
-    return {PropertyType::BOUNDARY_CONDITION, "transmission"};
-  return surface_metadata_.at({surface, type});
+  try {
+    if (surface_metadata_.count({surface, type}) == 0)
+      return {PropertyType::BOUNDARY_CONDITION, "transmission"};
+    return surface_metadata_.at({surface, type});
+  }
+  catch (const std::out_of_range e) {
```
gonuke, line 90: Doesn't the `count` check above make this already more robust? Can you add tests that fail before this change and pass afterwards?

gonuke: Also, could this `count` approach be used for volume  properties as well, instead of the exception?

gonuke:
Looking at the `Property` system in `constants.h` it seems that 
1. There is already a note to formally separate volume properties from surface properties, and
2. There may be room to define some kind of PROPERTY_NOT_FOUND constant

src/mesh_manager_interface.cpp
```diff
@@ -71,15 +71,25 @@ MeshManager::surface_has_property(MeshID surface, PropertyType type) const
 Property
 MeshManager::get_volume_property(MeshID volume, PropertyType type) const
 {
-  return volume_metadata_.at({volume, type});
+  try { 
+    return volume_metadata_.at({volume, type});
+  }
+  catch (const std::out_of_range e) {  
+    return VOID_MATERIAL;
```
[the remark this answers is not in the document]
Waqar-ukaea: Agreed. I have added a `PROPERTY_NOT_FOUND` constant as you suggested below.

src/mesh_manager_interface.cpp
Waqar-ukaea: Agreed on this one too. Added a similar `count` check for `get_volume_property()`

src/mesh_manager_interface.cpp
Waqar-ukaea: I'm not really certain what the best default for this would be so I'm going to leave it as is for now. Maybe @pshriwise  has a better idea of what we should do here.

gonuke: In theory, surfaces could get tagged with metadata for purposes other than boundary conditions, e.g. some variance reduction parameters, etc.  In such a case, a default that represents a boundary condition may not be the best default.

PR #113, opened 2025-06-05
Adding an fmt::formatter for the Vec3da type

pshriwise:
This adds a method that makes generating `fmt` messages with `Vec3da` objects much easier.


gonuke (approved):
LGTM

PR #116, opened 2025-06-05
Faster/simpler sidepair IDs for libMesh

pshriwise:
In setting up larger models with libMesh I found that, unsurprisingly, a ton of time was spent in `std::max_element` figuring out what the next available sidepair ID was. 

This replaces that approach by simply incrementing an internal ID value. It's not very sophisticated, but it's efficient.


pshriwise:
> The second conditional that is nested in the first conditional seems redundant and inconsistent with the overall logic flow. Am I missing something?

Hmm yeah that was probably left over from a refactor in the original addition for libMesh surface ray tracing. I don't think it's harmful but probably inefficient and can be remove. Will do so shortly. Thanks for the catch.

gonuke (approved):
This looks good to me - thanks @pshriwise

PR #121, opened 2025-06-09
Embree tetrahedral search callbacks

pshriwise:
This adds the set of callbacks needed for Embree to use the tetrahedron containment check function.

A new struct, `RTCElementRay` that supports discovery of a containing element with a zero-length array. Some high-level notes have been added to describe the process by which element searches will occur using these methods.

In isolation, these callbacks don't add any functionality to the library, but I feel it's a relatively coherent starting point before adding the code that includes the construction of and querying of Embree geometries and scenes.

### DIFF
--- a/docs/methods/element_search.rst
+++ b/docs/methods/element_search.rst
@@ -1,0 +1,44 @@
+
+
+Element Search
+==============
+
+The underlying data structure for element search used by Embree is a Bouding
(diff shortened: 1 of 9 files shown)


pshriwise:
Gonna go ahead and get this one merged in so we can move on to the next one.

PR #124, opened 2025-06-12
Adding tests for extended Embree structs

pshriwise:
This adds some basic unit tests for the various extended `struct` types defined in `ray.h`.

### DIFF
--- a/tests/test_ray_duals.cpp
+++ b/tests/test_ray_duals.cpp
@@ -1,0 +1,247 @@
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/matchers/catch_matchers_floating_point.hpp>
+
+#include "xdg/ray.h"
+#include "xdg/vec3da.h"
+
+using namespace xdg;
+
+TEST_CASE("Test RTCDualRay properties")
+{
+  RTCDualRay ray;
+
+  SECTION("Default constructor initializes values correctly")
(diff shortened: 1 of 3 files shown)


tests/test_ray_duals.cpp
```diff
@@ -0,0 +1,247 @@
...
+  SECTION("Default constructor initializes values correctly")
+  {
+    REQUIRE(ray.tnear == 0.0);
+    REQUIRE(ray.tfar == INFTYF);
+    REQUIRE(ray.mask == -1);
+  }
+
+  SECTION("set_org with array")
+  {
+    double org[3] = {1.0, 2.0, 3.0};
```
gonuke, line 22: Does it make sense to have a test that shows the difference between the single and double precision data?

gonuke (commented):
This all looks good to me. Just one question about a mixed precision test?

tests/test_ray_duals.cpp
pshriwise: Hmmmmm I don't really think so in this case. Unless we're looking to test something specific about difference between the values we're really just showing that they're fundamentally different types.

pshriwise:
> I assume the submodule update is at least benign, even if not directly related to this PR?

Benign, but inadvertent. I've reverted that change.

PR #131, opened 2025-06-25
MOAB direct access refactor

pshriwise:
This refactors the direct access objects used to query MOAB face and element data so that elements and faces use the same data structure to reference vertex coordinates. IMO it makes the process of vertex coordinate lookups for these types easier to follow.


pshriwise:
Gonna go ahead and merge this in as it's a refactor. I ran it against our current benchmarks as well and didn't observe a regression in performance.

PR #144, opened 2025-09-11
Simplify MOAB tag data retrieval

pshriwise:
This PR replaces direct calls to MOAB's `tag_get_data` methods. The primary motivation for this was the discovery in #143 that passing an empty to container to these methods results in a segfault within MOAB. Adding wrappers to perform a defensive check in a few places makes more sense than including them at a higher level, and at that point templating the methods also makes sense so we don't have a wrappers for each datatype we're gathering. There are a couple of methods that also enable us to get a single value for a single-handle query, which admittedly aren't necessary but it's quite convenient to be able to do:

```cpp
int dim = this->tag_data<int>(geometry_dimension_tag_, entity);
```

instead of 

```cpp 
int dim;
this->moab_interface()->tag_get_data(geometry_dimension_tag_, &entity, 1, &dim);
```

@Waqar-ukaea I know I know, I said I'd like to avoid over-templating, but I think it makes sense here I swear!

### DIFF
--- a/include/xdg/moab/mesh_manager.h
+++ b/include/xdg/moab/mesh_manager.h
@@ -114,12 +114,19 @@ public:
   }
 
   template<typename T>
-  std::vector<T> tag_data(moab::Tag tag, const std::vector<moab::EntityHandle>& entities) const {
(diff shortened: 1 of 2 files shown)


gonuke (approved):
Thanks @pshriwise

PR #149, opened 2025-09-24
XDG Configuration Object

pshriwise:
This adds `XDGConfig` with a singleton method `XDGConfig::config()` that allows one to set library parameters and perform runtime checks for what is enabled in the current XDG build. 

This object will also handle forwarding of parameters to library initialization as needed. Currently, this aids in setting the number of threads in the libMesh initialization -- needed primarily because libMesh will set the number of threads to zero by default, even when using an OpenMP threading model.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -21,7 +21,6 @@ XDGConfig::XDGConfig(int n_threads) {
   #endif
   }
 
-
   // Initialize libMesh if enabled
   #ifdef XDG_ENABLE_LIBMESH
   initialize_libmesh();
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -5,6 +5,8 @@
 #include "omp.h"
 #endif
 
+using namespace xdg;
+
 XDGConfig::XDGConfig(int n_threads) {
   // Set the number of threads
   if (n_threads != -1) {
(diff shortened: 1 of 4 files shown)


tests/test_overlap_check.cpp
```diff
@@ -32,7 +44,7 @@ TEST_CASE("Overlapping Volumes Test")
   REQUIRE(expected_overlaps == overlap_map.begin()->first);
 }
 
-TEST_CASE("Non-Overlapping Volumes Test")
+TEST_CASE_METHOD(OMP_SingleThreadFixture, "Non-Overlapping Volumes Test", "[single-thread]")
```
Waqar-ukaea, line 47: I'm not familiar with the idea but from what I can gather we define a struct which has `TEST_CASE_METHODS` associated with it rather than independent test cases? What is the reason to move towards this over independent `TEST_CASES`?

gonuke (changes requested):
A few thoughts/questions about intended behavior here @pshriwise  - thanks for this addition

src/config.cpp
```diff
@@ -0,0 +1,60 @@
...
+void XDGConfig::initialize_libraries() {
+#ifdef XDG_ENABLE_LIBMESH
+  // libmesh requires the program name, so at least one argument is needed
+  int argc = 1;
+  const std::string argv{"XDG"};
+  const char *argv_cstr = argv.c_str();
+  libmesh_init_ = std::make_unique<libMesh::LibMeshInit>(argc, &argv_cstr, 0, n_threads_);
+#endif
+
+  initialized_ = true;
```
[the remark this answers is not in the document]
pshriwise: Yep. It's mainly an indicator that this method has been called and isn't necessarily tied to anything happening.

src/config.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+  // if threads aren't manually specified,
+  // set using OpenMP
+  if (n_threads_ == -1) {
+  #ifdef XDG_HAVE_OPENMP
+    n_threads_ = omp_get_num_threads();
+  #else
+    n_threads_ = 1;
+  #endif
+  }
```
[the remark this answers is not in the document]
pshriwise: Totally agree. If I didn't have to make some additional changes to ensure that we can successfully set the number of threads before initializing, I would certainly have adopted this.

tests/test_overlap_check.cpp
pshriwise: Alrighty, we still need some kind of fixture/test setup that sets the number of threads to one but now can use the config object to do so instead of an OpenMP call. Let me know if you'd like to discuss the use of `TEST_CASE_METHOD` further.

pshriwise:
I've reworked a number of things here that allow configuration settings to be made without requiring an `XDGConfig::initialize` call. I think it makes downstream application code cleaner. Open to others' thoughts though!

src/config.cpp
```diff
@@ -0,0 +1,125 @@
...
+// External pointers that can be set by external applications
+const libMesh::LibMeshInit* external_libmesh_init {nullptr};
+const libMesh::Parallel::Communicator* external_libmesh_comm {nullptr};
+} // namespace config
+
+// Custom cleanup function that will be called at program exit
+// This ensures LibMeshInit is destroyed after all other global destructors
+static void cleanup_libmesh_at_exit() {
+  if (config::xdg_libmesh_init) {
+    config::xdg_libmesh_init.reset();
```
gonuke: Maybe my C++ knowledge is really getting antiquated, but if this is a pointer (`unique_ptr`), don't you need to dereference this as a pointer?
```suggestion
    config::xdg_libmesh_init->reset();
```

pshriwise:
Okay comments should be addressed here now. Thanks for the reviews @gonuke @Waqar-ukaea!

PR #151, opened 2025-10-08
Independent vertex contiguity

pshriwise:
Upon testing some larger benchmark model where the MOAB `Skinner` is used to generate an exterior surface of triangles to support partial domain tallies, I found that new triangles ended up in a new contiguous range of `EntityHandle`'s but used existing vertices on the existing tetrahedron for their connectivity.

As a result, an invalid contiguity block was passed to the `VertexData` objects in the `MOABDirectAccess` manager. In hindsight it makes sense that this would happen and in order to better support changing meshes, the `VertexData` object has been updated to handle its own contiguity ranges independent of the face and element data.

### DIFF
--- a/include/xdg/moab/direct_access.h
+++ b/include/xdg/moab/direct_access.h
@@ -45,7 +45,7 @@ public:
     return true;
   }
 
-  //! \brief Get the coordinates of a triangle as MOAB CartVect's
+  //! \brief Get the coordinates of a triangle as XDG Vertices
   inline std::array<xdg::Vertex, 3> get_mb_coords(const EntityHandle& tri) {
     auto [i0, i1, i2] = face_data_.get_connectivity_indices<3>(tri);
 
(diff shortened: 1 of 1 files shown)


include/xdg/moab/direct_access.h
```diff
@@ -266,20 +266,33 @@ class MBDirectAccess {
...
-    void set_coords(int idx, int i, xdg::Vertex& v) {
+    void set_coords(int i, xdg::Vertex& v) {
+      // determine the correct contiguous memory block index to use
+      int idx = 0;
+      auto fe = first_vertices[idx];
+      while(true) {
+        if (i - fe.first < fe.second) { break; }
+        idx++;
+        fe = first_vertices[idx];
+      }
```
gonuke, line 284: Is this the same?
```suggestion
      while(i - fe.first >= fe.second) {
        fe = first_vertices[++idx];
      }
```
(I have a personal bias against `while(true)` loops unless necessary... )

PR #178, opened 2025-11-12
Adding a ray batch query API 

### LINKED ISSUE #72: Array-based ray query interface
To more cleanly support event-based transport and GPU ray tracing queries, XDG should support array-based queries.

Waqar-ukaea:
This PR adds a set of methods to the XDG API that allow for launching batches of rays in one call. The main intention here being to plug into GPRT to perform GPU ray tracing at scale across the RT pipeline. 

At this stage, I have a working set of overloads for `GPRTRayTracer::point_in_volume()` and `GPRTRayTracer::ray_fire()` which can be used to call rays over large batches. 

The two unit tests `test_point_in_volume` and `test_ray_fire` have been extended to also test the batch variants of the methods but have been gated to only do so when GPRT is enabled (until I get a working implementation for embree too).

Three new miniapps have been added to `xdg/tools`. These are:
- `batch-ray-fire` - Provide a set of origins and directions and fire rays using the batch version of `xdg::ray_fire()`
- `batch-point-in-volume` - Provide a set of points and (optionally) directions to perform point in volume checks using the batch version of `xdg::point_in_volume()`
- `ray-benchmark` - Provide a `.h5m` model to perform a ray tracing throughput benchmark against that model (sample directions from unit sphere around origin provided)

The most interesting of the new miniapps is by far `ray-benchmark` as it allows for a direct performance comparison between our embree and GPRT ray tracing implementations. On the `jezebel.h5m` model with 40 million rays launched from the origin and intersecting the volume we see the following metrics.

**Embree:**
image

**GPRT:**
Screenshot from 2025-11-12 18-09-19

Which is a promising sign that we already have some performance increase. Hopefully coming PRs will increase this gap further.


Waqar-ukaea:
I've re-run my ray throughput benchmark miniapp with a larger model and I've also moved the timing region to be around the actual raygen launch so it is purely timing the ray tracing computation, rather than inadvertently also including the time taken to transfer ray buffers to device and hit buffers back to the host. 

I am running this on my local machine which contains an Nvidia RTX ada 2000 mobile (used by GPRT) and an 13th Gen Intel® Core™ i7-13850HX (used by Embree). Specs of the [RTX ada 2000](https://www.techpowerup.com/gpu-specs/rtx-2000-ada-generation.c4199) and the [i7-13850HX](https://www.intel.com/content/www/us/en/products/sku/232161/intel-core-i713850hx-processor-30m-cache-up-to-5-30-ghz/specifications.html):
<p align="center">
  image 1
  image 2
</p>

The model used along with a depiction (significantly smaller number of rays plotted) of the rays launched is shown in the image below:
ray_benchmark_tokamak_setup
### Benchmark parameters

| model_name         | Volume | No. of Elements | No. of Rays |
|--------------------|--------|-----------------|-------------|
| Paramak_tokamak.h5m| 2      | 280K            | 50M         |

### Ray tracer performance

**Note - GPRT results do not take into account device buffer IO. So comparison to Embree is a little disingenuous.**

| Ray Tracer backend     | Ray tracing Wall Time (sec) | Ray Tracing Throughput (ray/sec) | Speedup vs Embree |
|------------------------|-----------------------------|-----------------------------------|-------------------|
| Embree                 | 10.7318                     | 4.65905e+06                       | 1×                |
| GPRT (FP64)            | 0.762942                    | 6.55358e+07                       | ~14.1×            |
| GPRT (FP32) + RT cores | 0.0252805                   | 1.97781e+09                       | ~424.5×           |

So, a fairly significant speedup already seen in pure ray tracing throughput in moving from Embree to GPRT whilst maintaining the full mixed precision algorithm. However, if there is a possibility that we can move to full single precision ray tracing we can see a speedup of nearly 425x. And this isn't even a particularly powerful graphics-oriented chip and is a few years old now. An RTX 5090 (launched early January 2025) has around 10x the FP32 FLOPS as well as 7x the number of RT cores.

Waqar-ukaea:
Following on from the trouble I was having outlined in the last comment, I have opted for a different route to writing raydata on device which involves exposing the rayhit buffers and allowing the "external" application to write to them directly. This is actually the original approach I wanted to do. But I wasn't sure about which types to expose to the public facing API. But now I have a solution which seems to work okay. 

I define a new struct to hold device pointers for XDG's own double precision ray and hit structs. And this POD type is returned by `xdg->get_device_rayhit_buffers(N)`:
```cpp
struct DeviceRayHitBuffers {
  dblRay* rayDevPtr; // device pointer to ray buffers
  dblHit* hitDevPtr; // device pointer to hit buffers
  uint capacity = 0;
}
```
Once the external application has the device pointers, they can be passed to a compute shader and written to directly. The caveat here is that I have only tested this so far whilst using the same `GPRTContext` that XDG sets up. Of course a real world downstream application wont have this luxury and will likely not even be using Vulkan so I will need to figure out a way to get these vulkan device pointers into something meaningful that can be passed to another GPU API - Issue #182 for more detail on that. 

However, what this does mean is that I can more fairly compare the GPRT ray tracing against embree now. Since the device IO is now making use of the GPU rather than requiring an expensive host to device transfer, I will include it in the next benchmarks I run.

Waqar-ukaea:
Some new benchmarks after implementing the new approach to generate rays on device and some more consistent timing usage thanks to PR #148! 

### Benchmark parameters

| Model          | Volume | No. of Elements (faces) | No. of Rays | Location        |
|----------------|--------|-------------------------|-------------|-----------------|
| simple_tokamak | 2      | 280 080                 | 80 M        | (180, 250, -27) |

---

### Ray tracer performance (trace-only)

**Note:** These timings measure *only* the ray tracing kernel / API calls (no ray generation time).  
GPRT results still do not include any explicit device buffer I/O overhead, so direct comparison to Embree is somewhat optimistic in this metric.

| Ray Tracer backend | Ray Tracing Time (trace-only, s) | Ray Tracing Throughput (ray/s) | Speedup vs Embree |
|--------------------|-----------------------------------|---------------------------------|-------------------|
| Embree             | 17.3097                          | 4.62169e+06                     | 1×                |
| GPRT (FP64)        | 1.11261                          | 7.19033e+07                     | ~15.6×            |

---

### Ray tracer performance (end-to-end: generation + tracing)

These timings include both **random ray generation** and **ray tracing**, but *exclude* XDG initialisation and argument parsing.

| Ray Tracer backend | Generation + Tracing Time (s) | End-to-End Throughput (ray/s) | Speedup vs Embree |
|--------------------|-------------------------------|-------------------------------|-------------------|
| Embree             | 18.4366                       | 4.33921e+06                   | 1×                |
| GPRT (FP64)        | 1.96591                       | 4.06936e+07                   | ~9.4×             |

- **Embree**  
  - Random ray generation (CPU): `1.12685 s`  
  - Trace-only time: `17.3097 s`  
  - Generation + tracing: `18.4366 s` → `4.33921e+06 rays/s`

- **GPRT (FP64)**  
  - Random ray generation (compute shader): `0.853306 s`  
  - Trace-only time: `1.11261 s`  
  - Generation + tracing: `1.96591 s` → `4.06936e+07 rays/s`

---

### XDG initialisation and full wall-clock time

| Ray Tracer backend | XDG Init Time (s) | Full Wall-Clock Time (s, post-argparse) |
|--------------------|-------------------|------------------------------------------|
| Embree             | 0.310338          | 18.8146                                  |
| GPRT (FP64)        | 1.59102           | 3.55703                                  |

I'll need to do some refactoring before I can get a similar set of benchmarking in place for FP32 GPRT.

Waqar-ukaea:
The previous benchmarks have been against Embree running on a single thread which isn't exactly indicative of a real performance comparison. When running Embree on all of the threads available to me, the results on the GPU are much more modest: 
image

include/xdg/gprt/ray_tracer.h
```diff
@@ -100,9 +118,23 @@ class GPRTRayTracer : public RayTracer {
...
+    void check_rayhit_buffer_capacity(const size_t N) override;
+
+    // Method to expose device ray and hit buffers for external population
+    DeviceRayHitBuffers get_device_rayhit_buffers(const size_t N) override;
+
+    void pack_external_rays(void* origins_device_ptr, 
+                            void* directions_device_ptr,
+                            size_t num_rays) override;
+
+    GPRTContext context()
```
pshriwise: Ideally any methods working with or returning GPRT-specific information will be private in this class.

include/xdg/ray_tracing_interface.h
```diff
@@ -123,6 +167,110 @@ class RayTracer {
...
+   * @param[in] directions An array of Direction objects representing the directions of the rays
+   * @param[in] num_rays The number of rays to be processed in the batch
+   * @param[out] hitDistances An output array to store the computed intersection distances for each ray
+   * @param[out] surfaceIDs An output array to store the MeshIDs of the surfaces hit by each ray
+   * @param[in] dist_limit (optional) maximum distance to consider for intersections
+   * @param[in] orientation (optional) flag to consider whether Entering/Exiting hits should be rejected. Defaults to EXITING
+   * @param[in] exclude_primitives (optional) vector of surface element MeshIDs to exclude from intersection tests
+   * @return Void. Outputs stored in hitDistances and surfaceIDs arrays
+   */  
+  virtual void ray_fire(TreeID tree,
```
pshriwise: It looks like this assumes that all of the rays are fired against the same TLAS. In transport, particles are going to be in different cells/volumes and therefore will be firing against different TLAS. Is that possible to do in the same call I wonder?

src/gprt/ray_tracer.cpp
```diff
@@ -326,7 +342,159 @@ std::pair<double, MeshID> GPRTRayTracer::ray_fire(SurfaceTreeID tree,
...
-                
+
+void GPRTRayTracer::point_in_volume(TreeID tree,
+                                          const Position* points,
+                                          const size_t num_points,
+                                          uint8_t* results,
+                                          const Direction* directions,
+                                          std::vector<MeshID>* exclude_primitives)
+{
+  if (num_points == 0) return; // no work to do. Early exit
```
pshriwise: This line got me thinking about these queries in the larger context, connected to particle transport. Different particles will, potentially, require different geometric queries. Rather than resize buffers and re-pack ray data, do you think we might be able to indicate (for some rays) that there's no work to do and the ray query is a no-op for that ray?

pshriwise (changes requested):
I started in on this with a few lower level comments and ran into a few higher level questions that we might want to address before I continue to review, so I'm putting in a partial review now to start discussion on those points.

src/gprt/ray_tracer.cpp
```diff
@@ -222,19 +225,22 @@ bool GPRTRayTracer::point_in_volume(SurfaceTreeID tree,
...
 
+  const Direction defaultDir = Direction{1. / std::sqrt(2.0), 1. / std::sqrt(2.0), 0.0};
+
   // Use provided direction or if Direction == nulptr use default direction
   Direction directionUsed = (direction != nullptr) ? Direction{direction->x, direction->y, direction->z} 
-                            : Direction{1. / std::sqrt(2.0), 1. / std::sqrt(2.0), 0.0};
+                            : defaultDir;
+
+  // Catch directions with zero length
+  const double l2 = directionUsed.x*directionUsed.x
```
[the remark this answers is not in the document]
Waqar-ukaea: Nope this could be a separate PR, in principle it could happen in scalar queries too I think. I just happened to see it crop up when passing a large batch of queries.

pshriwise:
> @pshriwise When running openmc in event based mode, does the number of particles in flight ever change? i.e is there a need to change the number of rays to be launched in a single batch on the fly?

I think I may have answered this offline, but currently no the number of particles in flight doesn't change once transport begins -- no need for the ability to change the number of ray queries on the fly .

src/gprt/ray_tracer.cpp
```diff
@@ -164,8 +164,11 @@ GPRTRayTracer::create_surface_tree(const std::shared_ptr<MeshManager>& mesh_mana
     geom_data->normals = gprtBufferGetDevicePointer(normal_buffer);
     geom_data->primitive_refs = gprtBufferGetDevicePointer(primitive_refs_buffer);
     geom_data->num_faces = num_faces;
-    
-    gprtComputeLaunch(aabbPopulationProgram_, {num_faces, 1, 1}, {1, 1, 1}, *geom_data);
+
+    constexpr uint32_t threadsPerGroup = 64; // must match [numthreads(64,1,1)]
```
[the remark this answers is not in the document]
pshriwise: Sounds good. I'm okay with leaving this but let's make an issue so we remember to revisit the topic.

Waqar-ukaea:
Marking this as closed as it is no longer relevant with subsequent PRs which have been opened instead.

PR #205, opened 2026-02-27
Gprt vol tracing

Waqar-ukaea:
This PR adds all of the necessary setup to allow GPRT to perform ray tracing against volumetric (tet) meshes. 

Changes: 
-
- Added all required GPRT setup/shader plumbing to handle construction of BVHs on tet meshes as well as shaders to trace against those BVHs
- Added `GPRTRayTracer::find_element()` and updated `test_find_element` to test across both Embree and GPRT backends. **Note -** important to stress that this is a single ray dispatch version and an batch based dispatch will come in a later PR. Single dispatch versions do remain very useful for testing parity with Embree however.
- For device side AABB population I have implemented a slang generic (equivalent to C++ templating) which will be nice for future PRs that add support for elements with differing numbers of nodes.
- Restructured file structure for shaders to split triangle and tet shaders
- Made `plucker_tet_containment_test` cross compatible between slang and C++. In doing so I have opted to make use of Cramer's rule rather than matrix inversion since I was having trouble with matrices between host and device. GPRT's `double3x3` required implementing my own matrix inversion function and had a different row-column ordering compared to linalg's implementation.
- Some minor QOL changes across GPRTRayTracer

It looks like a much larger PR than it really is since this PR also makes some structural changes to the file layout across shaders. Namely it splits shaders into three .slang files `rt_common.slangh`, `triangle_rt_shaders.slang` and `tetrahedron_shaders.slang`. So around 🟩 **+180 additions** and  🟥 **−270 deletions** are solely from moving the existing shaders for surface ray tracing into the new files. 

Extra notes:
-
- This is not an implementation of the full volumetric tracking algorithm we have in place that allows us to walk elements 
- But rather just all of the RT side BVH setup and shader setup to do element point containment checks against the BVH in find_element()
- Since the PR for batch based array queries isn't in place this is just an initial attempt with each raygen shader launching a single ray.  

I guess in theory should we ever need to actually perform ray tracing operations against tets it also allows for that (would just need to define a new custom intersection shader for double precision) but as far as i know there is no use case for this required by OpenMC?


include/xdg/geometry/dp_math.h
```diff
@@ -8,18 +8,19 @@ functions for dot product, cross product, and absolute value. In C++ compilation
...
-#ifdef __SLANG__
+#if defined(__SLANG__) || defined(__SLANG_COMPILER__)
 
 // Slang compilation, map dp::vec3 -> double3
 namespace dp { 
   typedef double3 vec3;
 
+  static const double DBL_EPS = 2.2204460492503131e-016;
+  static const double PLUCKER_ZERO_TOL = 20.0 * DBL_EPS;
+  static const double INFTY = 1.7976931348623157e+308; // std::numeric_limits<double>::max() is not available in slang
```
pshriwise:
```suggestion
  static const double INFTY = 1.7976931348623157e+308;
```

include/xdg/geometry/plucker.h
```diff
@@ -160,6 +160,37 @@ inline PluckerIntersectionResult plucker_ray_tri_intersect(dp::vec3 vertices[3],
...
 }
 
+// Plücker containment test for tetrahedra. Returns true if the point is inside the tetrahedron defined by vertices v0, v1, v2, v3.
+inline bool plucker_tet_containment_test(const dp::vec3 point,
+                                  const dp::vec3 v0,
+                                  const dp::vec3 v1,
+                                  const dp::vec3 v2,
+                                  const dp::vec3 v3) {
+
+  // TODO - I've decided to use Cramer's rule here instead of matrix inversion to make it easier to implement a cross-compilable version of this function
```
pshriwise, line 170: In general this looks fine to me. In my branch for quads/hexes I've moved to a Plucker coordinate check against face and I might recommend we change to that broadly as it naturally extends to linear elements with an arbitrary number of faces, but I'm curious, what was incompatible in the previous version?

include/xdg/gprt/ray_tracer.h
```diff
@@ -114,25 +108,30 @@ class GPRTRayTracer : public RayTracer {
...
     GPRTBufferOf<int32_t> excludePrimitivesBuffer_; //<! Buffer for excluded primitives
     
     // Geometry Type and Instances
     std::vector<gprt::Instance> globalBlasInstances_; //<! List of every BLAS instance stored in this ray tracer
     GPRTGeomTypeOf<DPTriangleGeomData> trianglesGeomType_; //<! Geometry type for triangles
+    GPRTGeomTypeOf<DPTetrahedronGeomData> tetrahedraGeomType_; //<! Geometry type for tetrahedra (not currently supported)
 
     // Ray Generation parameters
-    uint32_t numRayTypes_ = 1; // <! Number of ray types. Allows multiple shaders to be set to the same geometery
+    uint32_t numRayTypes_ = 2; // <! Number of ray types. 0=surface, 1=volume
```
pshriwise: Let's create a new `enum` for this property if there isn't one already.

pshriwise: Oh, I see there are some constants in `shared_structs.h`. Let's apply `RT_NUM_RAY_TYPES` here.

include/xdg/gprt/rt_common.slangh
```diff
@@ -0,0 +1,77 @@
...
+    float3 fpAabbMax = float3(dpAabbMax + double3(FLT_EPSILON, FLT_EPSILON, FLT_EPSILON));
+
+    aabb[2 * primID] = fpAabbMin;
+    aabb[2 * primID + 1] = fpAabbMax;
+}
+
+
+bool orientation_cull(in double3 ray, in double3 normal, in xdg::HitOrientation orientation) {
+    if (orientation == xdg::HitOrientation::ANY) return false; // No culling
+    if (orientation == xdg::HitOrientation::EXITING) return dot(ray, normal) < 0.0; // Cull exiting rays
```
pshriwise: Probably not as important in the slang kernel (and it might even be optimized out), but it might make sense to store the dot product in a local variable so we don't potentially compute it twice.

src/gprt/ray_tracer.cpp
```diff
@@ -369,4 +435,39 @@ void GPRTRayTracer::check_ray_buffer_capacity(size_t N)
...
+  // Retrieve the hit from the dblHit buffer
+  gprtBufferMap(rayHitBuffers_.hit);
+  dblHit* hit = gprtBufferGetHostPointer(rayHitBuffers_.hit);
+  auto primitive_id = hit[0].primitive_id;
+  gprtBufferUnmap(rayHitBuffers_.hit); // required to sync buffer back on GPU? Maybe this second unmap isn't actually needed since we dont need to resyncrhonize after retrieving the data from device
+  
+  if (primitive_id == -1)
+    return ID_NONE;
+  else
+    return primitive_id;
```
pshriwise: Is this simplification accurate?

```suggestion
  return primitive_id;
```

src/gprt/tetrahedron_rt_shaders.slang
```diff
@@ -0,0 +1,87 @@
+#ifndef XDG_GPRT_TETRAHEDRON_RT_SHADERS_SLANG
+#define XDG_GPRT_TETRAHEDRON_RT_SHADERS_SLANG
+
+#include "../../include/xdg/gprt/rt_common.slangh"
+
+struct SolidRayFirePayload {
+    int primitiveid; // ID of the primitive the point is contained within
```
pshriwise: naming thing: I'd like us to try to be consistent in our naming style. How about `primitive_id`?

src/gprt/tetrahedron_rt_shaders.slang
```diff
@@ -0,0 +1,87 @@
...
+
+    TracePoint(world, RAY_FLAG_NONE, 0xff, xdg::RT_VOLUME_RAY_INDEX, xdg::RT_VOLUME_MISS_INDEX, pointDesc, payload);
+
+    // Store the ID of the containing element in the hit buffer for CPU readback
+    record.hit[rayID].primitive_id = payload.primitiveid; // we're reusing dblHit's primitive_id field to store the ID of the containing element since we don't need hit distance or surf_id for this shader
+}
+
+[shader("miss")]
+void tet_miss(inout SolidRayFirePayload payload) {
+    payload.primitiveid = -1; // Mark the ray as outside all tetrahedra
```
pshriwise: Can we use `ID_NONE` here instead?

src/gprt/triangle_rt_shaders.slang
```diff
@@ -0,0 +1,184 @@
...
+};
+
+[shader("closesthit")]
+void ray_fire_hit(uniform DPTriangleGeomData record, inout SurfaceRayFirePayload payload, in DPAttribute attr) {
+    // Distance from the ray origin to the hit point
+    uint hit_kind = HitKind();
+    uint rayID = DispatchRaysIndex().x;
+
+    // There is some logic for handling next volumes inside the h5m-reader which I could make use of too
+    // TODO : Should the dblHit struct return the next volume ID for the ray back to the host
```
pshriwise: Responding to this comment randomly: We could do but I don't see a reason to necessarily. As long as we know what volume we're in and what surface was hit it's a really cheap operation to find the next volume and in the case of particle transport we may not use it if a collision will occur before the particle reaches that surface.

src/gprt/triangle_rt_shaders.slang
```diff
@@ -0,0 +1,184 @@
...
+    rayDesc.Origin = float3(record.ray[rayID].origin);
+    rayDesc.Direction = normalize(float3(record.ray[rayID].direction));
+    rayDesc.TMin = float(record.ray[rayID].tMin); 
+    rayDesc.TMax = float(record.ray[rayID].tMax); 
+
+    SurfaceAccelerationStructure world = record.ray[rayID].volume_accel_surf;
+
+    // Pass the ray's origin and direction to the payload
+    payload.distance = -1.0f;
+    payload.surf_id = -1;
```
pshriwise: Another possible place to use `ID_NONE`.

src/gprt/triangle_rt_shaders.slang
```diff
@@ -0,0 +1,184 @@
...
+populate_tri_aabbs(uint3 DispatchThreadID: SV_DispatchThreadID, uniform DPTriangleGeomData record) {
+    uint primID = DispatchThreadID.x;
+    uint3 indices = record.index[primID];
+    uint vertexIndices[3] = {indices.x, indices.y, indices.z};
+    populate_aabb<3>(primID, record.vertex, record.aabbs, vertexIndices);
+}
+
+// ------------------------------------------------ CUSTOM INTERSECTION SHADERS ------------------------------------------------
+
+/* 1D ray generation intersection with a double precision triangle using the Plucker intersection algorithm*/
```
pshriwise: Sorry for the stray: The term "1D" threw me here for a sec. 

```suggestion
/* Single ray generation intersection with a double precision triangle using the Plucker intersection algorithm*/
```

pshriwise (changes requested):
A few thoughts and comments here. 

One larger question: Where is the box expansion occurring in the GPRTRayTracer? I see the expansion using `FLT_EPSILON` in `populate_aabb`, but for larger volumes this expansion may not be enough to guarantee we don't run into false negatives when traversing the tree.

include/xdg/geometry/plucker.h
Waqar-ukaea: It's been a while, but I think the main issue was related to row/column convention differences between the matrix API used by GPRT/Slang and the `linalg` API used on the C++ side. The old method (still in main) was doing this matrix multiplication to solve for our barycentric coords: 
```cpp
    // Solve T * [λ1, λ2, λ3] = rhs
    double3 lambda123 = mul(inverse(T),{rhs.x, rhs.y, rhs.z});
```
I initially tried to implement the same logic on the Slang side, but I got different barycentric coordinates on CPU vs GPU. My understanding is that the two math APIs were interpreting the matrix row/column ordering differently which was leading to differences when I would call `mul(...)`.

I switched to Cramer's rule to avoid the matrix inversion and multiplication path entirely. That keeps the shared Slang/C++ implementation to just dot and cross products, which is easier to keep consistent across both C++ and slang.

> I might recommend we change to that broadly as it naturally extends to linear elements with an arbitrary number of faces, but I'm curious, what was incompatible in the previous version?

Agreed, seems reasonable to move towards a more arbitrary method.

include/xdg/gprt/ray_tracer.h
Waqar-ukaea: > Oh, I see there are some constants in shared_structs.h. Let's apply RT_NUM_RAY_TYPES here.

Yeah, not sure what the best place for these would be since they're constants but will only ever be used inside of `GPRTRayTracer` and the slang shaders so it didn't feel appropriate to put them in `constants.h`. But open to suggestions on that front.

src/gprt/ray_tracer.cpp
Waqar-ukaea: Yep, I believe so

src/gprt/triangle_rt_shaders.slang
Waqar-ukaea: Makes sense. I'll remove this comment then.

src/gprt/triangle_rt_shaders.slang
Waqar-ukaea: Agreed it is a little misleading. I've changed the comment to this in the latest commit:
```cpp
// Custom FP64 Plucker ray-triangle intersection algorithm for each ray-triangle pair
[shader("intersection")]
void DPTrianglePluckerIntersection(uniform DPTriangleGeomData record)
{
```

Waqar-ukaea:
@pshriwise bb61bf667a625b0e50eebd804284da9dfd6e0449 addresses some of the comments you left. Most of the changes are self-explanatory, but i'll highlight one here:

I added a new `shared_constants.h` header so the shader code can use `xdg::ID_NONE` instead of hard-coded `-1` values as you suggested. `constants.h` currently pulls in C++-only dependencies so it cant be included directly in the Slang shader path so I made a new header `shared_constants.h` which doesn't contain any C++ specific stuff.

I also moved the enums from `shared_enums.h` into that header and removed the old `shared_enums.h`.

A couple of small syntax changes were needed to keep the header Slang-compatible: `constexpr` was replaced with `static const`, and brace initialization like `MeshID ID_NONE {-1}` was replaced with `MeshID ID_NONE = -1`.

Waqar-ukaea:
> One larger question: Where is the box expansion occurring in the GPRTRayTracer? I see the expansion using FLT_EPSILON in populate_aabb, but for larger volumes this expansion may not be enough to guarantee we don't run into false negatives when traversing the tree.

I see your point, this is missing from GPRT atm and we only do the `FLT_EPSILON` expansion you saw. I'll think about how best to handle this and get a separate PR up for it once I figure it out

include/xdg/shared_constants.h
```diff
@@ -0,0 +1,35 @@
+#ifndef XDG_SHARED_CONSTANTS_H
+#define XDG_SHARED_CONSTANTS_H
+
+namespace xdg {
```
pshriwise: As discussed live this week, probably best to avoid changing the infrastructure too drastically to accommodate slang. Sorry for the mislead here.

Waqar-ukaea: No problem! I'll revert back to what I had before

include/xdg/gprt/ray_tracer.h
```diff
@@ -114,25 +108,30 @@ class GPRTRayTracer : public RayTracer {
...
+    GPRTComputeOf<DPTetrahedronGeomData> aabbTetPopulationProgram_; //<! AABB population program for double precision rays against tetrahedron geometries
+
     // Buffers 
     gprtRayHit rayHitBuffers_;
     GPRTBufferOf<int32_t> excludePrimitivesBuffer_; //<! Buffer for excluded primitives
     
     // Geometry Type and Instances
     std::vector<gprt::Instance> globalBlasInstances_; //<! List of every BLAS instance stored in this ray tracer
     GPRTGeomTypeOf<DPTriangleGeomData> trianglesGeomType_; //<! Geometry type for triangles
+    GPRTGeomTypeOf<DPTetrahedronGeomData> tetrahedraGeomType_; //<! Geometry type for tetrahedra (not currently supported)
```
pshriwise, line 123: Not currently supported?

include/xdg/gprt/ray_tracer.h
```diff
@@ -114,25 +108,30 @@ class GPRTRayTracer : public RayTracer {
...
 
     // Ray Generation parameters
-    uint32_t numRayTypes_ = 1; // <! Number of ray types. Allows multiple shaders to be set to the same geometery
+    uint32_t numRayTypes_ = RT_NUM_RAY_TYPES; // <! Number of ray types. 0=surface, 1=volume
```
pshriwise, line 126: Let's remove the values in the comments here lest they change in the code itself and we forget to update the values here.

include/xdg/geometry/plucker.h
```diff
@@ -160,6 +160,37 @@ inline PluckerIntersectionResult plucker_ray_tri_intersect(dp::vec3 vertices[3],
...
+
+  const double inv_det = 1.0 / det;
+  const double lambda1 = dp::dot(rhs, dp::cross(e1, e2)) * inv_det;
+  const double lambda2 = dp::dot(e0, dp::cross(rhs, e2)) * inv_det;
+  const double lambda3 = dp::dot(e0, dp::cross(e1, rhs)) * inv_det;
+  const double lambda0 = 1.0 - (lambda1 + lambda2 + lambda3);
+
+  const double barycentric_min = -dp::PLUCKER_ZERO_TOL;
+  const double barycentric_max = 1.0 + dp::PLUCKER_ZERO_TOL;
+
```
pshriwise, line 186: It seems like these could be compile-time constants, no?

include/xdg/gprt/shared_structs.h
```diff
@@ -18,7 +26,8 @@ struct dblRay
   int32_t exclude_count;           // Number of excluded primitives
   xdg::HitOrientation hitOrientation;
   int volume_tree; // TreeID of the volume being queried
-  SurfaceAccelerationStructure volume_accel; // The volume accel 
+  SurfaceAccelerationStructure volume_accel_surf; // The volume accel for surface acceleration structure
+  SolidAccelerationStructure volume_accel_solid; // The volume accel for solid acceleration structure
```
pshriwise, line 30: The use of solid here feels a little off. What about "ElementAccelerationStrucutre"?

pshriwise (changes requested):
A few more thoughts as I look over this today.

include/xdg/gprt/shared_structs.h
Waqar-ukaea: I do agree but both `SurfaceAccelerationStructure` and `SolidAccelerationStructure` are types defined in GPRT here - https://github.com/gprt-org/GPRT/blob/master/gprt/gprt_builtins.slang 

So I would need to put in a PR to make the name change upstream to make that change

Waqar-ukaea: Or I suppose I could typedef them here locally in XDG to have that extra consistency with the naming scheme we have.

pshriwise: Ah, I missed that. Thanks for clarifying. `typedef`'s sound reasonable to me. I think it would help with readability and allow us to use consistent terminology. Let's try it on.

Waqar-ukaea:
> One larger question: Where is the box expansion occurring in the GPRTRayTracer? I see the expansion using FLT_EPSILON in populate_aabb, but for larger volumes this expansion may not be enough to guarantee we don't run into false negatives when traversing the tree.

This has now been addressed in #222 and I recommend that #222 PR is merged before this PR is so that I can rebase those changes properly into this branch first.

PR #209, opened 2026-03-19
Update submodules during CMake configuration

pshriwise:
This is a QOL upgrade for those installing XDG and unfamiliar with git. A new block in the CMake file borrowed from OpenMC (with some tweaks) now checks out the intended version of the submodules used in the project.


gonuke (changes requested):
This looks straightforward - just one little robustness question....

gonuke (changes requested):
One quick doublecheck on the catch2 logic & a comment typo

gonuke (approved):
LGTM - thanks @pshriwise

PR #215, opened 2026-04-01
Use face adjacencies to enter element from implicit complement

pshriwise:
This contains a set of changes to allow the `XDG::sgements` call to use mesh adjacencies when finding an entering intersection into the volumetric mesh. `XDG::segments` is the primary call used to separate particle tracks for superimposed tally meshes in particle transport.

Currently, a finite ray given to `segments` can undergo a glancing intersection with a face when attempting to enter/re-enter the mesh. The query performed at the intersection location + a small amount can result in a failure to find an element in these cases, causing the method to return early as a result as it becomes unclear how to proceed in determining if the ray intersects any additional elements along it's length. This can prevent subsequent ray segments volumetric beyond this glancing intersection from being recorded.

The updates here use a new `get_boundary_face_element` to enter the element adjacent to the face. Meshes and transport simulations that have previously raise the warning in `segments` regarding failure to find an element after an intersection no longer do in this scenario and tally results match other tracklength estimator schemes in OpenMC. Further testing is warranted, however -- particularly integrated testing in the repo itself. While a good amount of testing was done for superimposed meshes that partially covered the domain, I don't think the geometries were such that glancing intersections were likely. We can now improve on this in benchmark and integrated testing.

The new `get_boundary_face_element` method is named as such because for MOAB meshes only these adjacencies are established in the direct access manager.

# Alternatives

When the `find_element` call fails at the entering intersection location, fire another ray from the location + `TINY_BIT` under the assumption that the initial intersection discovered is a glancing intersection and repeat this process at the next intersection location should one be found closer than what distance remains for the finite ray. This option likely isn't as robust as entering an element via adjacencies nor is it as performant due to the (now) unecessary `find_element` call when entering a volumetric mesh element.

TODO:

- [x] Provide tally equivalence data
- [x] Add pathological test case

### DIFF
--- a/include/xdg/moab/direct_access.h
+++ b/include/xdg/moab/direct_access.h
@@ -135,7 +135,7 @@ private:
           std::vector<EntityHandle> verts;
           for (auto idx : o) verts.push_back(conn[idx]);
           Range adj_ents;
-          rval = mbi->get_adjacencies(verts.data(), verts.size(), 3, true, adj_ents);
+          rval = mbi->get_adjacencies(verts.data(), verts.size(), 3, false, adj_ents);
 
           // there be at most two adjacent elements for a given face
           if (adj_ents.size() > 2) {
--- a/include/xdg/moab/direct_access.h
+++ b/include/xdg/moab/direct_access.h
@@ -91,6 +91,10 @@ public:
     return element_adjacency_data_.get_element_adjacencies(element);
   }
 
+  inline EntityHandle get_boundary_face_element(const EntityHandle& face) const {
+    return boundary_face_adjacency_data_.get_boundary_face_element(face);
+  }
+
   const std::vector<std::vector<int>>& get_face_ordering(EntityType entity_type) const {
     return element_adjacency_data_.ordering.at(entity_type);
   }
@@ -217,7 +221,7 @@ private:
 
     template <int N>
     std::array<size_t, N>
-    get_connectivity_indices(const EntityHandle& e) {
+    get_connectivity_indices(const EntityHandle& e) const {
       // determine the correct contiguous block index to use
       int block_idx = 0;
       auto fe = first_elements[block_idx];
@@ -241,6 +245,43 @@ private:
       first_elements.clear();
       vconn.clear();
       entity_range.clear();
+    }
+  };
+
+  struct BoundaryFaceAdjacencyData {
+    std::unordered_map<EntityHandle, EntityHandle> boundary_face_to_element_;
+
+    void setup(Interface* mbi, const ConnectivityData& face_data) {
+      ErrorCode rval;
+
+      for (const auto& face : face_data.entity_range) {
+        auto conn = face_data.get_connectivity_indices<3>(face);
+        std::array<EntityHandle, 3> verts = {conn[0], conn[1], conn[2]};
+
+        Range adjacent_elements;
+        rval = mbi->get_adjacencies(verts.data(), verts.size(), 3, true, adjacent_elements);
+        MB_CHK_SET_ERR_CONT(rval, "Failed to get MOAB boundary face adjacencies");
+
+        if (adjacent_elements.size() > 2) {
+          throw std::runtime_error("Found more than two elements adjacent to a face");
+        }
+
+        if (adjacent_elements.size() == 1) {
+          boundary_face_to_element_[face] = adjacent_elements[0];
+        }
+      }
+    }
+
+    EntityHandle get_boundary_face_element(const EntityHandle& face) const {
+      auto it = boundary_face_to_element_.find(face);
+      if (it == boundary_face_to_element_.end()) {
+        return xdg::ID_NONE;
+      }
+      return it->second;
+    }
+
+    void clear() {
+      boundary_face_to_element_.clear();
     }
   };
 
@@ -310,6 +351,7 @@ private:
   ConnectivityData face_data_;
   ConnectivityData element_data_;
   AdjacencyData element_adjacency_data_;
+  BoundaryFaceAdjacencyData boundary_face_adjacency_data_;
   VertexData vertex_data_;
 
   public:
--- a/src/libmesh/mesh_manager.cpp
+++ b/src/libmesh/mesh_manager.cpp
@@ -137,19 +137,13 @@ LibMeshManager::face_connectivity(MeshID face) const {
   return conn;
 }
 
-std::vector<MeshID>
-LibMeshManager::get_face_elements(MeshID face) const {
+MeshID
+LibMeshManager::get_boundary_face_element(MeshID face) const {
   const auto& sp = sidepair(face);
-
-  std::vector<MeshID> elements;
-  elements.reserve(2);
   if (sp.first()) {
-    elements.push_back(sp.first()->id());
+    return sp.first()->id();
   }
-  if (sp.second()) {
-    elements.push_back(sp.second()->id());
-  }
-  return elements;
+  return ID_NONE;
 }
 
 MeshID LibMeshManager::create_volume() {
--- a/src/libmesh/mesh_manager.cpp
+++ b/src/libmesh/mesh_manager.cpp
@@ -137,6 +137,21 @@ LibMeshManager::face_connectivity(MeshID face) const {
   return conn;
 }
 
+std::vector<MeshID>
+LibMeshManager::get_face_elements(MeshID face) const {
+  const auto& sp = sidepair(face);
+
+  std::vector<MeshID> elements;
+  elements.reserve(2);
+  if (sp.first()) {
+    elements.push_back(sp.first()->id());
+  }
+  if (sp.second()) {
+    elements.push_back(sp.second()->id());
+  }
+  return elements;
+}
+
 MeshID LibMeshManager::create_volume() {
   MeshID next_volume_id = *std::max_element(volumes_.begin(), volumes_.end()) + 1;
   return next_volume_id;
--- a/src/xdg.cpp
+++ b/src/xdg.cpp
@@ -188,16 +188,12 @@ XDG::segments(const Position& start,
       // }
 
       // Get the element on the other side of the hit face using adjacencies
-      auto adjacent_element = mesh_manager()->get_face_elements(exclude_primitives.back());
-      if (adjacent_element.size() == 0) {
+      auto adjacent_element = mesh_manager()->get_boundary_face_element(exclude_primitives.back());
+      if (adjacent_element == ID_NONE) {
         warning("Ray fire hit surface {}, but no adjacent elements were found on the other side of the surface.", hit.second);
         return segments;
       }
-      if (adjacent_element.size() > 1) {
-        warning("Ray fire hit surface {}, but multiple adjacent elements were found on the other side of the surface. This likely indicates a problem with the mesh connectivity.", hit.second);
-        return segments;
-      }
-      current_element = adjacent_element[0];
+      current_element = adjacent_element;
       exclude_primitives.clear();
     }
     auto vol_segments = mesh_manager()->walk_elements(current_element, r, u, distance);
--- a/src/xdg.cpp
+++ b/src/xdg.cpp
@@ -163,19 +163,15 @@ XDG::segments(const Position& start,
     MeshID volume = ID_NONE;
     if (current_element == ID_NONE) {
       // fire a ray against the implicit complement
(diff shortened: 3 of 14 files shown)


include/xdg/moab/direct_access.h
```diff
@@ -244,6 +248,43 @@ class MBDirectAccess {
...
+        Range adjacent_elements;
+        rval = mbi->get_adjacencies(verts.data(), verts.size(), 3, true, adjacent_elements);
+        MB_CHK_SET_ERR_CONT(rval, "Failed to get MOAB boundary face adjacencies");
+
+        if (adjacent_elements.size() > 2) {
+          throw std::runtime_error("Found more than two elements adjacent to a face");
+        }
+
+        if (adjacent_elements.size() == 1) {
+          boundary_face_to_element_[face] = adjacent_elements[0];
```
Waqar-ukaea, line 270: Just double checking I understand what's happening here: 
- Loop over all faces:
   - Get face vertex handles (hardcoded for triangle faces atm)
   - Find the adjacent element which shares the face boundary defined by the vertices?

include/xdg/moab/direct_access.h
```diff
@@ -244,6 +248,43 @@ class MBDirectAccess {
...
+
+    void setup(Interface* mbi, const ConnectivityData& face_data) {
+      ErrorCode rval;
+
+      for (const auto& face : face_data.entity_range) {
+        auto conn = face_data.get_connectivity_indices<3>(face);
+        std::array<EntityHandle, 3> verts = {conn[0], conn[1], conn[2]};
+
+        Range adjacent_elements;
+        rval = mbi->get_adjacencies(verts.data(), verts.size(), 3, true, adjacent_elements);
```
Waqar-ukaea, line 262: From the MOAB for [`get_adjacencies()`](https://ftp.mcs.anl.gov/pub/fathom/moab-docs/classmoab_1_1Interface.html#a2cb0794670be30e983b8ab2660e338fe) 

Get the adjacencies for a vector of entities to entities of a specified dimension.

**Parameters**

- `from_entities`: Vector of `EntityHandle` values whose adjacencies are queried.
- `num_entities`: Number of entities in `from_entities`.
- `to_dimension`: Dimension of the desired adjacent entities.
- `create_if_missing`: If `true`, MOAB creates any missing entities of the specified dimension. This is only useful when `to_dimension < dim(*from_entities)`.
- `adj_entities`: STL vector to which adjacent entities are appended.
- `operation_type`: Either `INTERSECT` or `UNION`, indicating whether to return the intersection or union of the adjacency sets for `from_entities`.

We seem to be setting `create_if_missing=true` here but should we be? My understanding is that the element would already be a part of the mesh.

Regardless, the documentation suggests that it is only useful when going from higher dimension to lower (perhaps it doesn't do anything otherwise) but in our case we are doing 0D (vertices) to 3D (tet) so should this parameter be set to false?

src/libmesh/mesh_manager.cpp
```diff
@@ -137,6 +137,15 @@ LibMeshManager::face_connectivity(MeshID face) const {
...
 }
 
+MeshID
+LibMeshManager::get_boundary_face_element(MeshID face) const {
+  const auto& sp = sidepair(face);
+  if (sp.first()) {
+    return sp.first()->id();
+  }
+  return ID_NONE;
+}
```
Waqar-ukaea, line 147: Much nicer to do in libmesh compared to MOAB i see 😅

src/xdg.cpp
```diff
@@ -170,12 +172,22 @@ XDG::segments(const Position& start,
...
+      // // determine what element is on the other side of this surface
+      // current_element = find_element(r + u * TINY_BIT);
+
+      // if (current_element == ID_NONE) {
+      //   // warning("Ray fire hit surface {}, but could not find element on the other side of the surface.", hit.second);
+      //   return segments;
+      // }
+
+      // Get the element on the other side of the hit face using adjacencies
+      auto adjacent_element = mesh_manager()->get_boundary_face_element(exclude_primitives.back());
```
Waqar-ukaea: In this case, is excluding those hit faces from future ray_fires in the loop the intention here? Or are we more so just using the vector as a history of faces hit to return the latest hit to get the boundary element? 

If it's the latter, should we rename `exclude_primitives` to reflect this?

tests/test_libmesh.cpp
```diff
@@ -63,6 +64,28 @@ TEST_CASE("Test BVH Build Brick")
...
+      REQUIRE(element != ID_NONE);
+
+      const auto elem_conn = mesh_manager->element_connectivity(element);
+      for (const auto vertex : face_conn) {
+        REQUIRE(std::find(elem_conn.begin(), elem_conn.end(), vertex) != elem_conn.end());
+      }
+    }
+  }
+}
+
```
Waqar-ukaea: We've got a libmesh test and a meshmock test for returning the boundary face elements. Should we look at also testing moab? 

Worth having a `TEMPLATE_TEST_CASE` to reuse test logic between mesh backends?

Waqar-ukaea (commented):
Seems like a very nice change. Cuts out an extra `find_element()` call and feels like a much more robust way of determining the element on the other side of the surface. 

Couple minor points for now. But I want to take another look again tomorrow, so I'l hold off on approving for now.

pshriwise:
Okay this took me a little longer than it probably should have but I've added some tests here to ensure that the changes to the `segments` method using the boundary face-element mapping fixes the problems with glancing intersections. Fixed an issue with repeated zero-distance queries as well.

include/xdg/moab/direct_access.h
pshriwise: I misread your comment so ignore my previous response if you would.

As you pointed out, because all of the entities are present in the mesh, the value of this flag isn't really of consequence. I tend to leave it as `true` just to rule out the possibility that something inadvertently missing  in the mesh database could cause a problem.

include/xdg/moab/direct_access.h
pshriwise: Correct.

src/libmesh/mesh_manager.cpp
pshriwise: True! Though we had to build these sidepair objects ourselves. They come for free with MOAB. :)

src/xdg.cpp
pshriwise: More the latter, I'll take a crack at renaming it.

tests/test_libmesh.cpp
pshriwise: Good idea! I'll add one in.

tests/test_tracks.cpp
Waqar-ukaea: Unrelated to the changes you have made in this PR but should we remove this line as it pollutes the test output with these print statements?

tests/test_tracks.cpp
```diff
@@ -112,4 +122,156 @@ TEST_CASE("Test Random Internal Tracks")
...
+          std::vector<std::pair<MeshID, double>> track_segments;
+          REQUIRE_NOTHROW([&]() { track_segments = xdg->segments(start, end); }());
+
+          const double total_length = std::accumulate(
+              track_segments.begin(), track_segments.end(), 0.0,
+              [](double sum, const std::pair<MeshID, double>& segment) {
+                return sum + segment.second;
+              });
+
+          DYNAMIC_SECTION(fmt::format("Backend = {}, Face = {}, Vertex = {}", mesh_backend, face, vertex))
```
Waqar-ukaea: I have a question here about the intention of the nested dynamic section. My understanding is that by nesting this we end up with a situation where catch2 reports every failing case (mesh_backend, face, vertex) because each case is run in a separate execution path. So the failing test output looks something like this:
```text
Test Single-Tet Glancing Vertex Intersection Tracks - MOAB_Interface
  Backend = MOAB
  Backend = MOAB, Face = 1, Vertex = (0, 1, 0)

FAILED:
  REQUIRE( total_length == Catch::Approx(1.0).margin(1e-12) )
with expansion:
  0.0 == Approx( 1.0 )


Test Single-Tet Glancing Vertex Intersection Tracks - MOAB_Interface
  Backend = MOAB
  Backend = MOAB, Face = 1, Vertex = (1, 0, 0)

FAILED:
  REQUIRE( total_length == Catch::Approx(1.0).margin(1e-12) )
with expansion:
  0.0 == Approx( 1.0 )

Test Single-Tet Glancing Vertex Intersection Tracks - MOAB_Interface
  Backend = MOAB
  Backend = MOAB, Face = 2, Vertex = (1, 0, 0)

FAILED:
  REQUIRE( total_length == Catch::Approx(1.0).margin(1e-12) )
with expansion:
  0.0 == Approx( 1.0 )

... and so on

```
Instead the inner `DYNAMIC_SECTION` could be replaced with an `INFO` call like this: 
```cpp
INFO(fmt::format("Backend = {}, Face = {}, Vertex = {}", mesh_backend, face, vertex));
``` 
which would instead report the first failure per backend with a message detailing the information passed to `INFO`:

```text
Test Single-Tet Glancing Vertex Intersection Tracks - LibMesh_Interface
  Backend = LIBMESH

FAILED:
  REQUIRE( total_length == Catch::Approx(1.0).margin(1e-12) )
with expansion:
  0.0 == Approx( 1.0 )
with message:
  Backend = LIBMESH, Face = 1, Vertex = (0, 1, 0)
```
I do personally think the nested `DYNAMIC_SECTION` is fine if we do want explicit reporting of every failure but then I think the whole per case setup should probably go into the dynamic section too? 
```cpp

          const Position start = vertex - direction * half_length;
          const Position end = vertex + direction * half_length;

          std::vector<std::pair<MeshID, double>> track_segments;
          REQUIRE_NOTHROW([&]() { track_segments = xdg->segments(start, end); }());

          const double total_length = std::accumulate(
              track_segments.begin(), track_segments.end(), 0.0,
              [](double sum, const std::pair<MeshID, double>& segment) {
                return sum + segment.second;
              });
```
Otherwise if `xdg->segments()` somehow fails we dont see it in the detailed section to see which case caused the fail, we would just see the backend that failed.

tests/test_tracks.cpp
```diff
@@ -112,4 +122,156 @@ TEST_CASE("Test Random Internal Tracks")
...
+          [](double sum, const std::pair<MeshID, double>& segment) {
+            return sum + segment.second;
+          });
+      const double reverse_length = std::accumulate(
+          reverse_segments.begin(), reverse_segments.end(), 0.0,
+          [](double sum, const std::pair<MeshID, double>& segment) {
+            return sum + segment.second;
+          });
+
+      DYNAMIC_SECTION(fmt::format("Backend = {}, Vertex = {}", mesh_backend, vertex))
```
Waqar-ukaea: Similar comment here for nested dynamic sections

tests/test_tracks.cpp
```diff
@@ -112,4 +122,156 @@ TEST_CASE("Test Random Internal Tracks")
...
+      {
+        REQUIRE(forward_length == Catch::Approx(expected_length).epsilon(0.00001));
+        REQUIRE(reverse_length == Catch::Approx(expected_length).epsilon(0.00001));
+        REQUIRE(forward_length == Catch::Approx(reverse_length).epsilon(0.00001));
```
Waqar-ukaea: While attempting to force a failure in this dynamic section I noticed something strange. If we have exactly 4 assertions failed it results in ctest reporting these as skipped when they actually failed. I think this is because we set `SKIP_RETURN_CODE 4` in the `\tests\CMakeLists.txt`:
```
    catch_discover_tests(${test}
        PROPERTIES
        SKIP_RETURN_CODE 4
        TEST_PREFIX "${test}::"
        EXTRA_ARGS "--reporter;console;--out=/dev/stdout")
endforeach()
```

Waqar-ukaea: Seems like this is a known problem with catch2/ctest interaction potentially? 

https://github.com/catchorg/Catch2/issues/2873

Waqar-ukaea (commented):
Looks good, just have a couple comments on some of the tests added.

tests/test_tracks.cpp
pshriwise: I like the style of the `INFO` output above. Let's go with that. Updating now.

tests/test_tracks.cpp
pshriwise: Okay I have a solution for this, but it's involved. Let's defer for a separate PR. I'll create an issue.

Waqar-ukaea:
Latest changes look good to me, I'll go ahead and merge.

Issue #212, opened 2026-03-22
Add support for multiple blocks per sideset in the `LibMeshManager`


nuclearkevin:
When XDG is initialized with a libMesh unstructured mesh that has multiple subdomains which share an external sideset, the application throws a segmentation fault. 

I've tracked this down to `LibMeshManager::determine_surface_senses()`, more specifically [the following block of code](
https://github.com/xdg-org/xdg/blob/main/src/libmesh/mesh_manager.cpp#L345-L360):

```c++
int reference_block = mesh_id_to_sidepair_.at(surface_faces.at(0)).first()->subdomain_id();
surface_senses_[surface_id] = {reference_block, ID_NONE};

for (const auto &face : surface_faces) {
  auto& pair = mesh_id_to_sidepair_.at(face);
  // swap the element positions based on subdomain ID if needed
  if (pair.first()->subdomain_id() != reference_block) {
    pair.swap();
    // if we've swapped a nullptr into the first position, we have a problem
    if (pair.first() == nullptr) fatal_error("Attempting to swap nullptr to first face value");
  }
  // set the sense of the surface with respect to the other block to reverse
  if (pair.second() != nullptr)
    surface_senses_[surface_id] = {reference_block, pair.second()->subdomain_id()};
}
```

For the case where subdomains share an external sideset, `reference_block` is somewhat meaningless as there are multiple subdomains associated with the surface. The end result is that `pair.swap();` is called when `pair.second()` is `nullptr`. Interestingly, the `fatal_error` here doesn't catch this. Immediately after the pointers are swapped `pair.swap()` attempts to dereference `pair.first()` (which is now a `nullptr`) to loop over element faces, resulting in the segmentation fault.

This bug is fairly severe for Cardinal-wrapping XDG as it prevents the use of most meshes generated automatically with MOOSE's built-in set of `MeshGenerator` classes. Those generators add sidesets to the outside faces to apply neutronics boundary conditions (or other BCs), which often are references by elements in different blocks. A workaround that I've been employing in Cardinal at the present is to delete all sidesets in the mesh, but this obviously won't work in the long run when considering the addition of mesh-based geometry to Cardinal-OpenMC via XDG (where you'll want to apply boundary conditions).

Here's an example mesh which runs into this problem. Left are the unique blocks. Right is one of the sidesets containing elements with each block.

[...]

PR #216, opened 2026-04-01
Support for  libMesh sidesets

pshriwise:
This PR addresses #212 by splitting sidesets spanning multiple blocks into separate surfaces in the mesh topology understood by XDG (the libMesh sideset data is unmodified). New data structures are introduces to support a one-to-many mapping of sidesets to discovered interface sets. The previous merge method mainly validates the model and most of the work handling merging of sidesets into the model is now done in `create_surfaces_from_sidesets_and_interfaces`.

As this is a more complex workflow I've tried to provide clear comments throughout. There are a couple of local functors and locally defined type aliases that were introduced at the suggestion of Codex -- there's nothing wrong with it but it's not my typical style and I may rework some of that. 

**Key changes:**

**Refactoring and Clarification of Sideset/Interface Handling:**
- Replaced the `merge_sidesets_into_interfaces` method with a new `map_sidesets_to_discovered_interfaces` method, which validates and maps explicit sidesets to discovered interface sets, recording associations and ensuring all sideset faces are properly represented.

**Surface Creation and Metadata Management:**
- Refactored `create_surfaces_from_sidesets_and_interfaces` to:
  - Split explicit sidesets into multiple surfaces if they span multiple interface pairs.
  - Assign unique surface IDs to implicit interface surfaces not covered by sidesets.
  - Track the mapping of sidesets to created surfaces for accurate metadata assignment.

**New Data Structures in `LibMeshManager`:**

  - `sideset_interface_map_`: is the coarse association map.
It answers: “Which discovered interface pair or pairs does this sideset belong to?”
It stores only the interface-pair keys, so it is useful for validating coverage and for knowing whether a sideset is one-to-one or one-to-many with discovered interfaces.

  - `sideset_interface_face_map_`: is the detailed grouping map.
It answers: “Which specific faces from this sideset belong to each discovered interface pair?”
This is what lets surface creation split a single sideset into multiple surfaces when the sideset spans multiple discovered interface sets.

  - `sideset_surface_map_`: is the final output association.
It answers: “After surfaces are created, which surface IDs came from this sideset?”
This is mainly used in parse_metadata() so the sideset’s boundary-condition metadata can be copied onto every resulting surface.

**Code Cleanup:**
- Removed unused utility functions (`intersects_set`, `contains_set`) that are no longer needed.

**Header File Update:**
- Included `<map>` in `mesh_manager.h` to support new data structures.

### DIFF
--- a/src/libmesh/mesh_manager.cpp
+++ b/src/libmesh/mesh_manager.cpp
@@ -61,8 +61,8 @@ void LibMeshManager::init() {
   // subdomains/volumes)
   discover_surface_elements();
 
-  // merge sidesets into interfaces where possible
-  merge_sidesets_into_interfaces();
+  // map sidesets to discovered interfaces where possible
+  map_sidesets_to_discovered_interfaces();
 
   // create surfaces from sidesets and interfaces
   create_surfaces_from_sidesets_and_interfaces();
@@ -242,7 +242,7 @@ void LibMeshManager::discover_surface_elements() {
   volume_element_id_map_ = IDBlockMapping<MeshID>(volume_element_ids);
 }
 
-void LibMeshManager::merge_sidesets_into_interfaces() {
+void LibMeshManager::map_sidesets_to_discovered_interfaces() {
   // validate that each sideset face is already represented by one of the
   // discovered interface sets. Same-block internal faces may appear in a
   // sideset but are ignored here.
--- a/src/libmesh/mesh_manager.cpp
+++ b/src/libmesh/mesh_manager.cpp
@@ -264,6 +264,12 @@ void LibMeshManager::merge_sidesets_into_interfaces() {
 
   sideset_interface_map_.clear();
   sideset_interface_face_map_.clear();
+  // loop over discovered interface sets of faces and validate that each face in
+  // the explicit sidesets is represented in one of the discovered interface
+  // sets. As we go, we will also build a mapping of the explicit sidesets to
+  // the discovered interfaces that they overlap with and the faces they
+  // contribute to each interface for use in surface creation and metadata
+  // assignment in subsequent steps
   for (const auto& [sideset_id, sideset_elems] : sideset_face_map_) {
     if (sideset_elems.empty()) continue;
 
@@ -275,10 +281,25 @@ void LibMeshManager::merge_sidesets_into_interfaces() {
         continue;
       }
 
+      // build interface pair for this face using the subdomain IDs of the elements on either side.
+      // If there is no neighbor element, use ID_NONE for the second subdomain ID.
+      // Then resolve this pair against the discovered interface pairs to determine which interface set this face belongs to.
       InterfacePair face_pair {
           face.first()->subdomain_id(),
           face.second() ? face.second()->subdomain_id() : ID_NONE};
+
+      // If for some reason the face is internal to a subdomain, continue and do
+      // not associate it with an interface pair. This may occur in cases where
+      // the mesh has internal faces that are marked as sidesets, which is not
+      // ideal but we will allow it to pass through without error for now since
+      // it does not impact the mesh-based topology of the geometry.
+      if (face_pair.first == face_pair.second) {
+        continue;
+      }
+
       InterfacePair resolved_pair;
+      // If the interface pair cannot be located in a disocvered interface set,
+      // then this face is not properly represented by the mesh-based topology and the missing face count is incremented.
       if (!resolve_interface_pair(face_pair, resolved_pair)) {
         missing_face_count++;
         continue;
@@ -289,10 +310,17 @@ void LibMeshManager::merge_sidesets_into_interfaces() {
         continue;
       }
 
+      // If this face is properly represented by a discovered interface, then we
+      // will record the association between the sideset and the interface pair
+      // it belongs to, as well as the face's membership in that interface pair
+      // for use in surface creation and metadata assignment in subsequent steps
       sideset_interface_map_[sideset_id].insert(resolved_pair);
       sideset_interface_face_map_[sideset_id][resolved_pair].push_back(face_id);
     }
 
+    // If any of the faces in explicit sidesets are not represented by the
+    // discovered mesh-based topology, the number of missing faces is reported
+    // and an error is raised.
     if (missing_face_count > 0) {
       fatal_error(
           "Sideset {} has {} face(s) that are not represented by discovered interface sets",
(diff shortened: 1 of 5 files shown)


src/libmesh/mesh_manager.cpp
```diff
@@ -252,79 +242,211 @@ void LibMeshManager::discover_surface_elements() {
...
+      InterfacePair face_pair {
+          face.first()->subdomain_id(),
+          face.second() ? face.second()->subdomain_id() : ID_NONE};
+
+      // If for some reason the face is internal to a subdomain, continue and do
+      // not associate it with an interface pair. This may occur in cases where
+      // the mesh has internal faces that are marked as sidesets, which is not
+      // ideal but we will allow it to pass through without error for now since
+      // it does not impact the mesh-based topology of the geometry.
+      if (face_pair.first == face_pair.second) {
```
Waqar-ukaea, line 296: What is the reason someone might define their mesh that way with internal sidesets like that?

Should we explicitly report to the user that these sidesets are not considered with some kind of warning/message even if we don't explicitly error?

src/libmesh/mesh_manager.cpp
```diff
@@ -252,79 +242,211 @@ void LibMeshManager::discover_surface_elements() {
...
-    // if this is a defined sideset, it should match one of the pairs in the
-    // interface map. If it doesn't based on the current ordering of subdomains,
-    // swap the order
-    if(subdomain_interface_map_.count(subdomain_pair) == 0) {
-      subdomain_pair = {subdomain_pair.second, subdomain_pair.first};
+void LibMeshManager::map_sidesets_to_discovered_interfaces() {
+  // validate that each sideset face is already represented by one of the
+  // discovered interface sets. Same-block internal faces may appear in a
+  // sideset but are ignored here.
+  using InterfacePair = std::pair<MeshID, MeshID>;
```
Waqar-ukaea, line 249: Not a huge deal at all but since this alias is used in multiple methods and is kind of a `LibMeshManager` class specific concept is it worth having this alias moved to be defined once in the class declaration rather than per method it is used?

src/libmesh/mesh_manager.cpp
```diff
@@ -252,79 +242,211 @@ void LibMeshManager::discover_surface_elements() {
...
+  std::vector<MeshID> sideset_ids;
+  sideset_ids.reserve(sideset_interface_face_map_.size());
+  for (const auto& [sideset_id, _] : sideset_interface_face_map_) {
+    sideset_ids.push_back(sideset_id);
   }
+  std::sort(sideset_ids.begin(), sideset_ids.end());
+
+  // here we loop over all explicit sidesets in sorted order to ensure that we
+  // assign surface IDs in a deterministic way
+  for (auto sideset_id : sideset_ids) {
```
Waqar-ukaea: Super minor suggestion but we could explicitly name the local copy of sideset_ids vector as `sorted_sideset_ids` just to make the reasoning for the temporary vector a little clearer?

```suggestion
  std::vector<MeshID> sorted_sideset_ids;
  sorted_sideset_ids.reserve(sideset_interface_face_map_.size());
  for (const auto& [sideset_id, _] : sideset_interface_face_map_) {
    sorted_sideset_ids.push_back(sideset_id);
}
  std::sort(sorted_sideset_ids.begin(), sorted_sideset_ids.end());

  // here we loop over all explicit sidesets in sorted order to ensure that we
  // assign surface IDs in a deterministic way
  for (auto sideset_id : sorted_sideset_ids) {
```

src/libmesh/mesh_manager.cpp
```diff
@@ -252,79 +242,211 @@ void LibMeshManager::discover_surface_elements() {
...
+  // discovered subdomain interfaces to complete the topological definition of
+  // the mesh
+  std::vector<InterfacePair> interface_pairs;
+  interface_pairs.reserve(subdomain_interface_map_.size());
+  for (const auto& [pair, _] : subdomain_interface_map_) {
+    interface_pairs.push_back(pair);
+  }
+  std::sort(interface_pairs.begin(), interface_pairs.end());
+
+  for (const auto& pair : interface_pairs) {
```
Waqar-ukaea: Similar to the previous suggestion name as `sorted_` for extra clarity? But again not really anything particularly required. 


```suggestion
  std::vector<InterfacePair> sorted_interface_pairs;
  sorted_interface_pairs.reserve(subdomain_interface_map_.size());
  for (const auto& [pair, _] : subdomain_interface_map_) {
    sorted_interface_pairs.push_back(pair);
  }
  std::sort(sorted_interface_pairs.begin(), sorted_interface_pairs.end());

  for (const auto& pair : sorted_interface_pairs) {
```

src/libmesh/mesh_manager.cpp
```diff
@@ -252,79 +242,211 @@ void LibMeshManager::discover_surface_elements() {
...
+  //    interface pair, assigning metadata from the sideset to the surface as we
+  //    go. If an explicit sideset is only associated with one interface pair,
+  //    we will reuse the sideset ID for the surface since there is no risk of
+  //    ID conflicts in this case, otherwise we will assign a new surface ID.
+  // 2. loop over discovered interface pairs and create surfaces for any pairs
+  //    that are not fully covered by explicit sidesets, assigning any remaining
+  //    metadata from associated sidesets to the surface  as we go. Since these
+  //    surfaces are not associated with explicit sidesets, we will assign new
+  //    surface IDs to all of these surfaces to avoid any potential ID
+  //    conflicts.
```
Waqar-ukaea, line 347: The comments here mention that this method assigns metadata but i'm not sure it does?

Waqar-ukaea (commented):
Overall looks like a nice set of changes to address the issue. A couple questions attached to specific lines and one more general one here:

Is it possible when meshing a libmesh mesh to have faces which belong to multiple sidesets? In `map_sidesets_to_discovered_interfaces()` we are looping through sidesets and add the `face_id` of a resolved `SidePair` to the `sideset_interface_face_map` but if the face belongs to multiple sidesets will it get added to that map multiple times?

src/libmesh/mesh_manager.cpp
pshriwise: > What is the reason someone might define their mesh that way with internal sidesets like that?

It's useful to know what faces are at the interface between blocks. Some applications (likely Griffin in the case of MOOSE) probably use them to compute surface quantities between reactor assemblies for example.



> Should we explicitly report to the user that these sidesets are not considered with some kind of warning/message even if we don't explicitly error?

Based on the code we use now they should still be respected I believe, though they may be split into multiple XDG surfaces if they span multiple blocks. Let me know if you mean something else though.

src/libmesh/mesh_manager.cpp
pshriwise: So, I initially agreed and I actually implemented this but I realized that with the existence of the `SidePair` struct it becomes a little unclear what we're using as a persistent reference for faces. The placement of this type alias makes it pretty clear that these objects are local to the method and won't be accessed elsewhere, which I like.

src/libmesh/mesh_manager.cpp
pshriwise: Hrmm we do update a new internal structure that allows us to reliably apply metadata from a multi-block sideset to any resulting surfaces in XDG in a one-to-many mapping, but we don't actually apply it within this method. I'll adjust the wording accordingly.

src/libmesh/mesh_manager.cpp
Waqar-ukaea: Gotcha. Seems like this is the best approach then.

src/libmesh/mesh_manager.cpp
nuclearkevin: > What is the reason someone might define their mesh that way with internal sidesets like that?

Like Patrick said, these are useful when computing surface values inside a mesh. An example would be computing interface current between regions in a transport solver (which would be a nice capability to eventually support in OpenMC+XDG at some point). A more standard use is post-processing for diffusive fluxes / flowrates.

nuclearkevin:
> Is it possible when meshing a libmesh mesh to have faces which belong to multiple sidesets?

Per the Exodus specification (and therefore the libMesh spec), elements can only belong to a single set of their dimensionality. For a 3D mesh, that means that 3D elements (TET4, HEX8, PRISM6, etc.) can only belong to a single block at a time, a 2D element can only belong to a single sideset (though, a 3D element may have multiple associated sidesets as they contain multiple 2D face elements), and a 1D element can only belong to a single edgeset.

Waqar-ukaea:
I see, thanks @nuclearkevin!

src/libmesh/mesh_manager.cpp
Waqar-ukaea: > Based on the code we use now they should still be respected I believe, though they may be split into multiple XDG surfaces if they span multiple blocks. Let me know if you mean something else though.

That clears things up for me, I just got myself a little mixed up on the logic here

nuclearkevin (approved):
Looks good to me, thanks for patching this so quickly!

Waqar-ukaea (approved):

PR #217, opened 2026-04-01
Refactor overlap checking

Waqar-ukaea:
I threw this small PR together to close #214 since it seemed like a relatively quick thing to finish off. 

I've tried cloning a fresh install with `XDG_BUILD_TOOLS=OFF` and it looks like the submodule for indicators isn't pulled in (in the same way that argparse isnt too). 

@pshriwise not sure if you were thinking of a more in depth refactor here for the overlap checking code but I've gone ahead and just moved the headers and source code directly into `\tools` and then changed all the required CMake stuff too.


tests/CMakeLists.txt
```diff
@@ -34,8 +34,8 @@ if (XDG_ENABLE_MOAB AND XDG_ENABLE_LIBMESH)
 endif()
 
 # placing this last as it's a slow test
-if (XDG_ENABLE_MOAB)
```
pshriwise, line 37: Yeah these tests are definitely long, but I wouldn't mind exercising both mesh managers here. If the tests are taking too long maybe we can address it in a separate PR.

vendor/GPRT
pshriwise, line 1: Should there be changes to the GPRT submodule for this PR?

tests/CMakeLists.txt
pshriwise: Let's leave this as-is for now, but I'd like to refactor further so we don't include the progress bars in `overlap.cpp`/`overlap.h`.

Can you remove the comments though? They're more discussion points than guideposts as to what/why the code is the way it is.

pshriwise (changes requested):
Thanks for tackling this @Waqar-ukaea! Just a couple of comments. I'd still like to exercise this code in testing. Ideally only the tool executable itself would require the `indicators` library. This a great intermediate step to that and we can take it the rest of the way in a follow-on PR.

tests/CMakeLists.txt
Waqar-ukaea: I've gone ahead and removed the comments. 

> but I wouldn't mind exercising both mesh managers here

To be clear, you mean run these tests with both mesh managers via the `TEMPLATE_TEST_CASE_METHOD` pattern we have established for other tests which want to run with multiple backends? 

That would be pretty easy to setup I think but would need me to also generate the equivalent `.exo` files for the test files used in them as right now I think we only have the `.h5m` ones. So perhaps best suited for a later PR?

pshriwise: Yeah, totally good with that waiting for a separate PR. Mind making an issue to track it though?

Waqar-ukaea: Yep, will do.

vendor/GPRT
Waqar-ukaea: This should be reverted now

tests/CMakeLists.txt
Waqar-ukaea: > Let's leave this as-is for now, but I'd like to refactor further so we don't include the progress bars in overlap.cpp/overlap.h.

Do you want me to address this refactor in a further PR or within this one?

pshriwise: Up to you! Slight preference from me for a separate PR as this set of changes is pretty clear at this point.

Waqar-ukaea: Yeah i think it makes more sense to be a separate PR too. In which case I'd say if you're happy, maybe merge this one?

PR #222, opened 2026-04-21
Ensure BLAS only built once per surface with GPRT

Waqar-ukaea:
# Summary
This PR addresses #168 and refactors `GPRTRayTracer::create_surface_tree()` so that surface geometry setup and BLAS construction are cached per surface in the same spirit as the existing Embree implementation.

A new private method, `GPRTRayTracer::register_surface()`, now encapsulates the one-time setup for a surface:

- gather vertices, connectivity, primitive ids, and normals
- create the corresponding GPRT device buffers
- initialize the `DPTriangleGeomData` record to point at those buffers

In refactoring the logic for BLAS creation, I have also made sure that the correct bounding box bump is applied and passed to the compute kernel which constructs bounding boxes.

# Key Changes
- `register_surface()` is only called on the first visit to a surface.
- New struct to hold GPRT buffers for better lifetime management - previously host side handles weren't stored so couldn't be destroyed later
- Surface geometry and backing buffers are cached and recovered on the second visit through `surface_to_geometry_map_`and `surface_buffers_map_`.
- Surface AABBs are now populated using the maximum bounding-box bump of the two parent volumes for that surface.
- BLAS construction is now performed once per surface and cached in `surface_to_blas_map_`.
- On the second visit from the other parent volume, the implementation reuses the cached geometry and BLAS rather than rebuilding them, and only updates the per-volume tree metadata and TLAS instance list.
- Ran particle sim with a multi-volume model and results remained consistent after refactor (also agrees with Embree)

### DIFF
--- a/include/xdg/geometry/plucker.h
+++ b/include/xdg/geometry/plucker.h
@@ -25,8 +25,6 @@ struct PluckerIntersectionResult {
   double t = 0.0;      // Distance along the ray to the intersection point
 };
 
-constexpr PluckerIntersectionResult EXIT_EARLY = {false, 0.0};
-
 /* Function to return the vertex with the lowest coordinates. To force the same
   ray-edge computation, the Plücker test needs to use consistent edge
   representation. This would be more simple with MOAB handles instead of
@@ -81,7 +79,7 @@ inline PluckerIntersectionResult plucker_ray_tri_intersect(dp::vec3 vertices[3],
   // If orientation is set, confirm that sign of plucker_coordinate indicate
   // correct orientation of intersection
   if (useOrientation && orientation * plucker_coord0 > 0) {
-    return EXIT_EARLY;
+    return {false, 0.0};
   }
 
   // Determine the value of the second Plucker coordinate from edge 1
@@ -92,13 +90,13 @@ inline PluckerIntersectionResult plucker_ray_tri_intersect(dp::vec3 vertices[3],
   // correct orientation of intersection
   if (useOrientation) {
     if (orientation * plucker_coord1 > 0) {
-      return EXIT_EARLY;
+      return {false, 0.0};
     }
     // If the orientation is not specified, all plucker_coords must be the same
     // sign or zero.
   } else if ((0.0 < plucker_coord0 && 0.0 > plucker_coord1) ||
              (0.0 > plucker_coord0 && 0.0 < plucker_coord1)) {
-    return EXIT_EARLY;
+    return {false, 0.0};
   }
 
   // Determine the value of the third Plucker coordinate from edge 2
@@ -109,7 +107,7 @@ inline PluckerIntersectionResult plucker_ray_tri_intersect(dp::vec3 vertices[3],
   // correct orientation of intersection
   if (useOrientation) {
     if (orientation * plucker_coord2 > 0) {
-      return EXIT_EARLY;
+      return {false, 0.0};
     }
     // If the orientation is not specified, all plucker_coords must be the same
     // sign or zero.
@@ -117,12 +115,12 @@ inline PluckerIntersectionResult plucker_ray_tri_intersect(dp::vec3 vertices[3],
              (0.0 > plucker_coord1 && 0.0 < plucker_coord2) ||
              (0.0 < plucker_coord0 && 0.0 > plucker_coord2) ||
              (0.0 > plucker_coord0 && 0.0 < plucker_coord2)) {
-    return EXIT_EARLY;
+    return {false, 0.0};
   }
 
   // check for coplanar case to avoid dividing by zero
   if (0.0 == plucker_coord0 && 0.0 == plucker_coord1 && 0.0 == plucker_coord2) {
-    return EXIT_EARLY;
+    return {false, 0.0};
   }
 
   // get the distance to intersection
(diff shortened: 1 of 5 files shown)


Waqar-ukaea:
> In refactoring the logic for BLAS creation, I have also made sure that the correct bounding box bump is applied and passed to the compute kernel which constructs bounding boxes

The new cross checking test in #225 exposed the current bounding box expansion with GPRT as being insufficient and leading to missed rays that embree does not miss. So this PR being merged will also fix that problem too.

src/gprt/ray_tracer.cpp
```diff
@@ -203,10 +171,79 @@ GPRTRayTracer::create_surface_tree(const std::shared_ptr<MeshManager>& mesh_mana
...
+  // Get storage for indices
+  auto indices = mesh_manager->get_surface_connectivity(surface_id);
+  std::vector<uint3> ui3Indices;
+  ui3Indices.reserve(indices.size() / 3);
+  for (size_t i = 0; i < indices.size(); i += 3) {
+    ui3Indices.emplace_back(indices[i], indices[i + 1], indices[i + 2]);
+  }
+
+  // Get storage for normals
+  std::vector<double3> normals;
```
pshriwise, line 205: Can you remind me why we're storing the face normals explicitly on the surface buffer?

src/gprt/ray_tracer.cpp
```diff
@@ -331,13 +368,22 @@ std::pair<double, MeshID> GPRTRayTracer::ray_fire(SurfaceTreeID tree,
 void GPRTRayTracer::create_global_surface_tree()
 {
   // Create a TLAS (Top-Level Acceleration Structure) for all the volumes
-  auto globalBuffer = gprtDeviceBufferCreate<gprt::Instance>(context_, globalBlasInstances_.size(), globalBlasInstances_.data());
-  GPRTAccel global_accel = gprtInstanceAccelCreate(context_, globalBlasInstances_.size(), globalBuffer);
+  std::vector<gprt::Instance> globalBlasInstances;
```
pshriwise, line 371: This is a local variable, do we need to track these instances so they can be cleaned up later or does that happen automatically?

include/xdg/geometry/plucker.h
```diff
@@ -25,7 +25,7 @@ struct PluckerIntersectionResult {
   double t = 0.0;      // Distance along the ray to the intersection point
 };
 
-constexpr PluckerIntersectionResult EXIT_EARLY = {false, 0.0};
+static constexpr PluckerIntersectionResult EXIT_EARLY = {false, 0.0};
```
pshriwise, line 28: Why the addition of `static` here?

pshriwise (changes requested):
Just a couple of questions here, but it generally all looks good to me.

src/gprt/ray_tracer.cpp
Waqar-ukaea: I need the normal available in the instersection shader on device to do sense adjustment of the normal at intersection time. I suppose I could recover the normal directly from the vertices on device in the double precision intersection shader with a cross product though.

include/xdg/geometry/plucker.h
Waqar-ukaea: It was to suppress a slang warning:

[  3%] Generating dbl_deviceCode.spv
/home/waqar/xdg/include/xdg/geometry/plucker.h(28): warning 39019: 'EXIT_EARLY' is implicitly a global shader parameter, not a global variable. If a global variable is intended, add the 'static' modifier. If a uniform shader parameter is intended, add the 'uniform' modifier to silence this warning.
constexpr PluckerIntersectionResult EXIT_EARLY = {false, 0.0};
                                    ^~~~~~~~~~
I didn't really read much into it beyond the warning suggesting it should have the static keyword applied. But it does still compile without static so I can remove it if you like

src/gprt/ray_tracer.cpp
Waqar-ukaea: That I think is fine actually. There is no `gprtInstanceDestroy()` method because the instances themselves are very simple structs: 
```cpp
struct Instance {
  float3x4 transform;

  uint32_t instanceCustomIndex : 24;
  uint32_t mask : 8;
  uint32_t geometryIndex : 24;
  uint32_t flags : 8;

  uint64_t accelAddress;
};
```
so the instances themselves don't store any resources which need to be cleaned up later on.

src/gprt/ray_tracer.cpp
pshriwise: Yeah I tend to compute these on-the-fly to save memory. Are you okay with doing the same here? We can address in a separate PR.

src/gprt/ray_tracer.cpp
pshriwise: Excellent. Thanks for looking into this!

include/xdg/geometry/plucker.h
pshriwise: The fewer warnings the better, so I'd say leave it as-is. Thanks!

src/gprt/ray_tracer.cpp
Waqar-ukaea: Yep, makes sense to me.

Waqar-ukaea: Separate PR is good for me too

pshriwise (approved):
Thanks @Waqar-ukaea! On to the next one :)

PR #224, opened 2026-05-01
Resolve unintended linking to stale XDG library.

pshriwise:
I noticed recently that when compiling with libMesh enabled, the test executables will link to a installed version of the XDG library if it is located in the same location as the libMesh dependency. Seeing as it's a common practice to install optional/source-built software in the same location in one's user space, this PR to resolve this by adding the `PUBLIC`/`PRIVATE` specifiers for `target_link_libraries` CMake command.

This would often cause problems when switching between versions of XDG containing different symbols or enabling/disabling libMesh support.

Some changes were required elsewhere to make this happen. Because some of the tests include `mesh_managers.h` resulting in dependency headers being included directly. So the test executables are now linked directly to the dependencies and link to the build version of the XDG library.

The `mesh_manager.h` include has been removed from the various tools and as a result these mini-apps need only to be linked to the XDG build library. Linking only to the XDG shared binary desired behavior IMO and the way these tools are built is now representative of this approach.


Waqar-ukaea (changes requested):
Happy with the changes here and feels like a step in the right direction. However I think setting the compile definitions for `XDG_ENABLE_GPRT` and `XDG_ENABLE_MOAB` is resulting in tests being skipped even when those libraries are enabled. 

I _think_ the compile definitions are no longer being propagated into the tests when these are set to `PRIVATE` which I note we don't do for `XDG_ENABLE_LIBMESH` and those tests seem to be working fine with no skips. 

Annoyingly github isnt letting me add comments to the particular lines :/ and it just keeps saying "an internal error has occured". 

The lines I am suggesting changes to are `CMakeLists.txt:308`:
```CMake
target_compile_definitions(xdg PRIVATE XDG_ENABLE_MOAB)
```
Should be changed to:
```CMake
target_compile_definitions(xdg PUBLIC XDG_ENABLE_MOAB)
```
 And `CMakeLists.txt:320`:
```CMake
target_compile_definitions(xdg PRIVATE XDG_ENABLE_GPRT)
```
Should be changed to:
```CMake
target_compile_definitions(xdg PUBLIC XDG_ENABLE_GPRT)
```

pshriwise:
> Happy with the changes here and feels like a step in the right direction. However I think setting the compile definitions for `XDG_ENABLE_GPRT` and `XDG_ENABLE_MOAB` is resulting in tests being skipped even when those libraries are enabled.
> 
> I _think_ the compile definitions are no longer being propagated into the tests when these are set to `PRIVATE` which I note we don't do for `XDG_ENABLE_LIBMESH` and those tests seem to be working fine with no skips.
> 
> Annoyingly github isnt letting me add comments to the particular lines :/ and it just keeps saying "an internal error has occured".
> 
> The lines I am suggesting changes to are `CMakeLists.txt:308`:
> 
> ```cmake
> target_compile_definitions(xdg PRIVATE XDG_ENABLE_MOAB)
> ```
> 
> Should be changed to:
> 
> ```cmake
> target_compile_definitions(xdg PUBLIC XDG_ENABLE_MOAB)
> ```
> 
> And `CMakeLists.txt:320`:
> 
> ```cmake
> target_compile_definitions(xdg PRIVATE XDG_ENABLE_GPRT)
> ```
> 
> Should be changed to:
> 
> ```cmake
> target_compile_definitions(xdg PUBLIC XDG_ENABLE_GPRT)
> ```

Thank you! Good catch. I've made the suggested changes here and confirmed that no tests are skipped with all mesh and raytracing backends enabled.

tests/CMakeLists.txt
Waqar-ukaea:
```suggestion
    target_link_libraries(test_overlap_check PRIVATE indicators::indicators)
```

CI seems to be failing because of an instance of `target_link_libraries` which doesn't use a scope keyword. I think the suggestion above should fix.

PR #225, opened 2026-05-08
Introduce new cross checks for ray tracing backends

Waqar-ukaea:
This PR is to introduce some more automated testing - namely cross checking results between ray tracing backends. 

The idea is to create multiple XDG instances with the same mesh manager and file loaded but with different ray tracer backends (depending on what is enabled at configuration time) and cross check the results between each of them to ensure the standard queries in the XDG interface provide consistent results. 

I am hoping to write this in such a way that it can be easily extended to include the cuBQL backend too. 

**Also renamed test_cross_check to hopefully make more sense with the new test file added**


tests/util.h
```diff
@@ -33,36 +33,59 @@ struct StringMaker<std::pair<MeshTag, RayTag>> {
...
+    case xdg::RTLibrary::EMBREE:
+    #ifdef XDG_ENABLE_EMBREE
+      return true;
+    #else
+      return false;
+    #endif
+
+    case xdg::RTLibrary::GPRT:
+    #ifdef XDG_ENABLE_GPRT
+      return system_has_vk_device();
```
pshriwise, line 47: This is interesting. A bit tangent, but should this runtime check also be in the constructor of the GPRTRayTracer?

tests/test_ray_tracer_cross_check.cpp
```diff
@@ -0,0 +1,94 @@
...
+} // namespace
+
+TEST_CASE("Test Pincell RT libraries Cross-Check ray_fire queries", "[moab][rayfire][cross-check]")
+{
+  const auto rt_cases = make_rt_cases("pincell.h5m");
+  if (rt_cases.size() < 2) {
+    SKIP("Fewer than two ray tracing backends are available; skipping cross-check.");
+  }
+
+  std::vector<Direction> directions {
```
pshriwise: Any reason not to use randomly selected directions in this test?

pshriwise: I think we could also perform more queries 100 if not 1000 I'd say.

pshriwise (changes requested):
Small thoughts from me here. Thanks for knocking this out @waqar-ukaea!

tests/test_ray_tracer_cross_check.cpp
Waqar-ukaea: > Any reason not to use randomly selected directions in this test?

I did consider it but thought hardcoded would be more consistent for a test. But now that I think about it we could do RNG with a fixed seed if we want that consistency. 

> I think we could also perform more queries 100 if not 1000 I'd say.

I'd hesitate to do that with this particular code pathway since it is exercising a single ray being launched with GPRT (and cuBQL eventually) which may end up being quite slow. Though once batch queries are in for GPRT (and similarly for cuBQL) absolutely agree with tuning up the number of queries.

tests/util.h
Waqar-ukaea: Hmmm. That might would probably be a good idea actually yeah. And I should probably extend it to check explicitly if the `vk_device` available has the necessary ray tracing extensions enabled.

tests/test_ray_tracer_cross_check.cpp
```diff
@@ -0,0 +1,94 @@
...
+      for (size_t i = 1; i < rt_cases.size(); ++i) {
+        const auto& candidate = rt_cases[i];
+        const auto candidate_hit = candidate.xdg->ray_fire(volume, origin, direction);
+
+        // Capture should print the name of the backend when test fails
+        CAPTURE(volume, direction, reference_case.name, candidate.name,
+                reference_hit.first, reference_hit.second, candidate_hit.first, candidate_hit.second);
+        REQUIRE(candidate_hit.second == reference_hit.second);
+        if (reference_hit.second != ID_NONE) {
+          REQUIRE_THAT(candidate_hit.first, Catch::Matchers::WithinAbs(reference_hit.first, 1e-6));
```
Waqar-ukaea, line 82: So my idea here was that we loop through all of the ray tracers and we take the first in the list as our "reference data" because it would likely be embree. But when there are three ray tracers enabled do you think there is a need to cross checking between the 2nd and 3rd? Or is cross checking between GPRT & embree and then cuBQL & embree enough?

tests/test_ray_tracer_cross_check.cpp
pshriwise: Even 100 queries would be too slow for a unit test?

tests/test_ray_tracer_cross_check.cpp
pshriwise: I think using the first as a reference is a good approach. I think we should be able to get them all to agree as long as we're using the same triangle intersection test.

tests/util.h
pshriwise: We could do that here or in a separate PR. I'll leave it up to you!

tests/test_ray_tracer_cross_check.cpp
Waqar-ukaea: I'm not actually sure tbh. I'll play around with it and if it's not too slow I'll go ahead and push the changes

Waqar-ukaea:
From slack: 

> [11:06 AM] Regarding your suggested changes to increase the number of queries up to 1000 (+ random directions) and it pushes up the test time to about 2 seconds with debug symbols on my machine so not actually as bad as what I had assumed.

> [11:07 AM] However bumping up the number of rays actually does expose a new bug in that I am seeing the test fail inconsistently despite using a fixed seed for the random directions

> [11:08 AM] So looking into what the problem is here now :confused:

> [11:21 AM] Nevermind I think I figured it out. On the branch in #225 bouding box expansion for GPRT is still only FLT_EPSILON and I think the extra number of rays is exposing that as unsuitable (edited) 

> [11:22 AM] Just tested with a temporary branch of #225 merged to the head of #222 (the PR which remove BLAS duplication per surface and fixes bounding box expansion for GPRT) and the test in #225 is consistently passing now

> [11:25 AM] So #222 is a blocker to ensuring the cross check against GPRT in #225 is consistently passing.

But I still wouldn't mind if #225 is merged first just to help with cuBQL development. GPRT failing the cross check wont be caught in CI since we aren't testing GPRT builds of XDG in the CI currently.

tests/util.h
Waqar-ukaea: I think I will make this change in a separate PR so will also open an issue to track that

pshriwise (approved):
Looks good! Thank you @Waqar-ukaea!

PR #227, opened 2026-05-10
Adding example

### LINKED ISSUE #87: Initial set of examples.
It's probably time for some C++ (and eventually Python) examples of how to use XDG.

magnoxemo:
closes #87 

I am opening to this PR to test if I am breaking anything or not since, I haven't build xdg locally yet.


magnoxemo:
@pshriwise I think this is ready for a review.

examples/ray_segments.cpp
```diff
@@ -0,0 +1,65 @@
...
+#elif defined(XDG_ENABLE_LIBMESH)
+    auto mesh_manager = std::make_shared<xdg::LibMeshManager>();
+    mesh_manager->load_file("brick.exo"); // Provide the path to your mesh file.
+#endif
+
+    mesh_manager->init();
+
+    // by default, we are using Embree as a ray tracer
+    auto xdg_instance = std::make_shared<xdg::XDG>(mesh_manager);
+    xdg_instance->prepare_raytracer();
```
Waqar-ukaea: Whilst this is a valid way to setup xdg and indeed what the constructor expects, the use of the compiler preprocessor guards can be avoided if you allow xdg to set up its mesh manager internally. Instead of using the constructor, the XDG factory method can be used to just specify the underlying mesh and ray tracing libraries you wish to create XDG with: 

```suggestion
// Read in your mesh file  
// Do whatever logic you need to decide whether you want to use xdg with MOAB or libmesh 
// Use xdg factory method XDG::create() - if no libraries specified it defaults to MOAB + embree
if (use_moab) {
   std::shared_ptr<XDG> xdg = XDG::create(xdg::MeshLibrary::MOAB, xdg::RTLibrary::EMBREE);
else {
   std::shared_ptr<XDG> xdg = XDG::create(xdg::MeshLibrary::LIBMESH, xdg::RTLibrary::EMBREE);
}

// Then we can recover the mesh manager object abstractly without needing to specify the derived mesh manager classes (and thus removing the need for the pre-compile guards)
const auto& mesh_manager = xdg->mesh_manager();
mesh_manager->load_file(filename);
mesh_manager->init();

xdg->prepare_raytracer();
```
The miniapps in `/tools` have some other examples of xdg instantiation too.

magnoxemo:
I just noticed that one of the example in tools have used argument parser. Maybe it will be better to take advantage of that? Let me know if you don't like the changes.

examples/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# =============================================================================
+# Examples for how to use XDG C++ APIs
+# =============================================================================
+
+function(xdg_add_examples name source)
```
pshriwise: Personal preference here: Until we have more than one example I'd prefer that we don't wrap all of this in a function. I generally prefer avoiding CMake functions because the meaning of local vs. global state is very easy to get confused.

Also, we _shouldn't_ need to link directly to the underlying mesh libraries, linking to the XDG target should suffice for an example like this. Same for some of the other libraries like `fmt` unless they're used in the example code. (I didn't catch any direct uses of `fmt` below).

examples/ray_segments.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "xdg/xdg.h"
+#include "xdg/constants.h"
+#include "xdg/mesh_managers.h"
+#include "argparse/argparse.hpp"
+#include "xdg/error.h"
+
+int main(int argc, char* argv[])
+{
+
+    argparse::ArgumentParser args("An example of how to setup XDG's C++ API  to calculate "
```
pshriwise: I appreciate the effort here, but for examples I'd probably prefer that we avoid any dependencies we don't need to reduce confusion regarding what's required for XDG. So no `ArgumentParser` would be my vote.

examples/CMakeLists.txt
pshriwise: Re `fmt` and my other comment: For clarity, let's try to leave `fmt` out of the equation if possible.

pshriwise (changes requested):
A few thoughts from me as well.

To give the example some purpose, I wonder if a Monte Carlo volume estimation using the XDG methods would be interesting.

magnoxemo:
Agreed. I will update this and make it a volume estimation example.

examples/element_volume_estimation.cpp
```diff
@@ -0,0 +1,52 @@
...
+        std::cerr << "Usage: " << argv[0] << " <mesh_file>\n";
+        return 1;
+    }
+    const std::string mesh_file = argv[1];
+
+    std::shared_ptr<xdg::XDG> xdg = xdg::XDG::create();
+    const auto& mesh_manager = xdg->mesh_manager();
+    mesh_manager->load_file(mesh_file);
+    mesh_manager->init();
+    xdg->prepare_raytracer();
```
Waqar-ukaea, line 19: Since you're relying on default behaviour for the xdg instance creation this will **always** create a `MOABMeshManager` under the hood so maybe some acknowledgement that this example should only ever be passed `.h5m` files? As it would fail to load an exodus file.

Waqar-ukaea: @pshriwise with this in mind, do you think it would ever be useful for a factory method that is passed a mesh filename and uses the file extension to determine whether or not to create an XDG instance with the MOAB backend vs libmesh backend?

pshriwise: Yeah I was thinking the same but it becomes tricky when MOAB is built with the ability to read exodus files as well and in the absence of geometry we should be able to use them for tallying (I think). 

In that case the file extension makes the mesh manager selection ambiguous. Maybe the preference for exodus would be to use libMesh if available and attempt to fall back to MOAB.

Waqar-ukaea: Oh I didnt realise that MOAB had the ability to read in exodus files I had just assumed we had a clean separation of: 
- `.exo` --> libmesh
- `.h5m` --> MOAB 

In which case my previous comment may be wrong too about this example currently failing to load an exodus file? 
>Maybe the preference for exodus would be to use libMesh if available and attempt to fall back to MOAB.

I think that is a reasonable compromise because presumably if you were trying to use an exodus file its more likely you want libmesh right? Although I'm not sure what the preference becomes once we also have an MFEM backend working too.

pshriwise: For clarity, let's punt on this for now and require that the filename ends with `.h5m`. Let's also explicitly request the `MOABMeshManager` by providing the `xdg::MeshLibrary::MOAB` argument in the example here.

CMakeLists.txt
```diff
@@ -323,7 +332,7 @@ if (XDG_ENABLE_GPRT)
   target_link_options(xdg PRIVATE -Wl,--unresolved-symbols=ignore-in-shared-libs)
 endif()
 
-target_link_libraries(xdg PRIVATE fmt::fmt)
+target_link_libraries(xdg PUBLIC fmt::fmt)
```
pshriwise, line 335: Can you remind me why this is necessary again?

magnoxemo: The issue was that `fmt` was linked privately to `xdg`. As a result, when a library or executable linked against `xdg`, the `fmt` dependency was not propagated to it. That caused the linker error we were seeing at that time.

PR #230, opened 2026-05-12
Extend vk probe to check for required Vulkan RT extensions

### LINKED ISSUE #229: Adding the check for Vulkan capable device directly to GPRTRayTracer
Inside of `test/vulkan_probe.h` there is a function `system_has_vk_device()` that is used to probe (at runtime) whether the current system has a Vulkan capable device available. 

I think I originally wrote this essentially as as quick stop to make sure that even if GPRT was enabled at configuration time an additional check at runtime was performed to skip generation of any GPRT tests if there is no vulkan ready device available. But @pshriwise rightly pointed out that this should probably live inside of the `GPRTRayTracer()` constructor too. 

It is also probably worth extending the checks to also query which Vulkan extensions are available on that device too. As there will be some GPUs which may have some vulkan support but not the full set of extensions needed for Vulkan ray tracing via GPRT.

Waqar-ukaea:
PR to close out #229 (more information enclosed in issue) and add GPRT to the CI build matrix. GPRT testing will of course still be skipped since there is no gpu available on the runners.


Waqar-ukaea:
Playing around with setting up a CI matrix for ray tracing backends in [1f1aa89](https://github.com/xdg-org/xdg/pull/230/commits/1f1aa89779be02ede928d79680ccb4dc60a317f9) but I may just go ahead and revert to how it was before

Waqar-ukaea:
CI failing because GPRT is being built as a static library inside the runner which is causing it to fail in linking some NVIDIA library stuff (namely DLSS I believe). So the fix here is to build GPRT as a shared library and update the CMakeLists.txt to properly handle that too.

Waqar-ukaea:
Extending the CI matrix to also build GPRT is proving to be quite painful in all honesty. I feel like I am jumping through all sorts of hacky hoops just to get it to compile on the runner. Starting to question whether or not its even worth it at this point.

pshriwise:
> Extending the CI matrix to also build GPRT is proving to be quite painful in all honesty. I feel like I am jumping through all sorts of hacky hoops just to get it to compile on the runner. Starting to question whether or not its even worth it at this point.

Okay. Let's not let it hold this up then. The intent is for it to be a simple improvment. Let's remove the CI changes and tackle it in a separate effort.

Waqar-ukaea:
> Okay. Let's not let it hold this up then. The intent is for it to be a simple improvment. Let's remove the CI changes and tackle it in a separate effort.

I wrote that comment just before I pushed one last change yesterday thinking it would fail to compile again but it actually did compile and shows the tests skipping as they should, so I might actually be there already. Whether or not you feel the changes I have made to the different CMakeLists.txt is reasonable I guess is a different question.

```
Total Test time (real) =   7.93 sec

The following tests did not run:
	 12 - test_ray_fire::Ray Fire on MeshMock (per-backend sections) - GPRT_Raytracer (Skipped)
	 14 - test_point_in_volume::Point-in-volume on MeshMock - GPRT_Raytracer (Skipped)
	 33 - test_xdg_interface::XDG Factory Creation - std::pair<MOAB_Interface, GPRT_Raytracer> (Skipped)
	 35 - test_xdg_interface::XDG Factory Creation - std::pair<LibMesh_Interface, GPRT_Raytracer> (Skipped)
	 37 - test_xdg_interface::XDG Constructor - std::pair<MOAB_Interface, GPRT_Raytracer> (Skipped)
	 39 - test_xdg_interface::XDG Constructor - std::pair<LibMesh_Interface, GPRT_Raytracer> (Skipped)
	 41 - test_xdg_interface::XDG Query API - std::pair<MOAB_Interface, GPRT_Raytracer> (Skipped)
	 43 - test_xdg_interface::XDG Query API - std::pair<LibMesh_Interface, GPRT_Raytracer> (Skipped)
	 56 - test_moab::Test BVH Build - GPRT_Raytracer (Skipped)
	 58 - test_moab::Test Ray Fire MOAB (all built backends) - GPRT_Raytracer (Skipped)
```
Finally the CI is still failing because GPRT isn't respecting its own CMake setting `GPRT_BUILD_SHARED`. When this variable is set to on it should build GPRT as a shared library but this is not the case with GPRT still building a static library. I have a PR open to address this - https://github.com/gprt-org/GPRT/pull/84 

So I think you're right, for now the best option will be to close this PR without the CI changes and update the CI at a later date.

Waqar-ukaea:
This branch should now be reverted back to before any of the CI changes were in place and is probably ready to merge if you are happy @pshriwise

pshriwise:
Minor conflict to address here now @Waqar-ukaea.

Waqar-ukaea:
@pshriwise should be good now

pshriwise (approved):
Looks good from here. Thanks!

PR #233, opened 2026-05-14
Add docker file

### LINKED ISSUE #231: Add a Docker file for xdg
A Docker container can be useful for quickly testing new changes in XDG before pushing to GitHub. It can help folks to isolate their system from different versions of libMesh and MOAB, which may also be used as dependencies by other frameworks.

magnoxemo:
closes #231

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,17 +18,10 @@ RUN apt-get install --yes \
         mpich \
         libmpich-dev \
         libtirpc-dev \
-        python3 \
-        python3-pip \
-        python3-packaging \
-        python3-jinja2 \
-        python3-yaml \
-        python3-pkgconfig \
         curl \
         nano \
         libtbb-dev \
-        libglfw3-dev \
-        libgl1-mesa-dev
+        libembree-dev
 
 # compile libmesh from source and install it
 RUN git clone --recurse-submodules https://github.com/libMesh/libmesh.git /XDG_TEST_SYSTEM/libmesh
@@ -54,19 +47,6 @@ RUN make install
 
 ENV MOAB_INSTALL_PATH=/XDG_TEST_SYSTEM/moab_install_dir
 
-# build Embree from source
-RUN git clone https://github.com/RenderKit/embree.git /XDG_TEST_SYSTEM/embree
-WORKDIR /XDG_TEST_SYSTEM/embree/build
-RUN cmake .. \
-        -DCMAKE_BUILD_TYPE=Release \
-        -DCMAKE_INSTALL_PREFIX=/XDG_TEST_SYSTEM/embree_install_dir \
-        -DEMBREE_ISPC_SUPPORT=OFF \
-        -DEMBREE_TUTORIALS=OFF \
-        -DEMBREE_STATIC_LIB=OFF
-RUN make -j$(nproc)
-RUN make install
-
-ENV EMBREE_INSTALL_PATH=/XDG_TEST_SYSTEM/embree_install_dir
 
 # build XDG
 RUN git clone --recurse-submodules -j$(nproc) https://github.com/xdg-org/xdg.git /XDG_TEST_SYSTEM/xdg
@@ -78,12 +58,6 @@ RUN cmake .. \
         -DXDG_ENABLE_MOAB=ON \
         -DMOAB_DIR=${MOAB_INSTALL_PATH} \
         -DXDG_ENABLE_LIBMESH=ON \
-        -DLIBMESH_DIR=${LIBMESH_INSTALL_PATH} \
-        -DXDG_LINK_MPI=ON \
-        -DXDG_ENABLE_EMBREE=ON \
-        -DCMAKE_PREFIX_PATH=${EMBREE_INSTALL_PATH} \
-        -DCMAKE_CXX_FLAGS="-I/${EMBREE_INSTALL_PATH}/include" \
-        -DXDG_BUILD_TESTS=ON \
-        -DXDG_BUILD_TOOLS=ON
+        -DLIBMESH_DIR=${LIBMESH_INSTALL_PATH}
 RUN make -j$(nproc)
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,8 @@ RUN apt-get install --yes \
         curl \
         nano \
         libtbb-dev \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        -DENABLE_HDF5=ON \
+        -DHDF5_ROOT=/usr \
+        -DBLAS_LIBRARIES=/usr/lib/x86_64-linux-gnu/libopenblas.so \
+        -DBUILD_SHARED_LIBS=ON
+RUN make -j$(nproc)
+RUN make install
+
+ENV MOAB_INSTALL_PATH=/XDG_TEST_SYSTEM/moab_install_dir
+
+# build Embree from source
```
pshriwise: Any reason not to rely on the apt package for Embree and save some build time? 

```bash
$ sudo apt install libembree-dev
```

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        -DCMAKE_C_COMPILER=mpicc \
+        -DCMAKE_CXX_COMPILER=mpicxx \
+        -DXDG_ENABLE_MOAB=ON \
+        -DMOAB_DIR=${MOAB_INSTALL_PATH} \
+        -DXDG_ENABLE_LIBMESH=ON \
+        -DLIBMESH_DIR=${LIBMESH_INSTALL_PATH} \
+        -DXDG_LINK_MPI=ON \
+        -DXDG_ENABLE_EMBREE=ON \
+        -DCMAKE_PREFIX_PATH=${EMBREE_INSTALL_PATH} \
+        -DCMAKE_CXX_FLAGS="-I/${EMBREE_INSTALL_PATH}/include" \
```
pshriwise: Is this line necessary? If so, we should look into it.

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        -DXDG_BUILD_TESTS=ON \
+        -DXDG_BUILD_TOOLS=ON
```
pshriwise: These are the default values already right?

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        libopenblas-dev \
+        mpich \
+        libmpich-dev \
+        libtirpc-dev \
+        python3 \
+        python3-pip \
+        python3-packaging \
+        python3-jinja2 \
+        python3-yaml \
+        python3-pkgconfig \
```
pshriwise: Do we really need all of this? We aren't installing `pymoab`...

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        curl \
+        nano \
+        libtbb-dev \
+        libglfw3-dev \
+        libgl1-mesa-dev
```
pshriwise: I don't think we'll need these either?

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        libtirpc-dev \
+        python3 \
+        python3-pip \
+        python3-packaging \
+        python3-jinja2 \
+        python3-yaml \
+        python3-pkgconfig \
+        curl \
+        nano \
+        libtbb-dev \
```
pshriwise: If we use `libembree-dev` see below, this should get installed with it I believe.

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        mpich \
+        libmpich-dev \
+        libtirpc-dev \
+        python3 \
+        python3-pip \
+        python3-packaging \
+        python3-jinja2 \
+        python3-yaml \
+        python3-pkgconfig \
+        curl \
```
pshriwise, line 23: Remove as well?

pshriwise (changes requested):
Looks good all-in-all @magnoxemo. Thanks! Just some thoughts here on thinning it out and maybe simplifying things.

Dockerfile
magnoxemo: Apologies. I just didn't know that. I will do that.

magnoxemo:
@pshriwise I have tested this locally and image builds nicely with this new recipe.

Dockerfile
pshriwise: Any reason to keep this now that we've included `libembree-dev`?

magnoxemo: Removed it

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+WORKDIR /XDG_TEST_SYSTEM/xdg/build
+RUN cmake .. \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_C_COMPILER=mpicc \
+        -DCMAKE_CXX_COMPILER=mpicxx \
+        -DXDG_ENABLE_MOAB=ON \
+        -DMOAB_DIR=${MOAB_INSTALL_PATH} \
+        -DXDG_ENABLE_LIBMESH=ON \
+        -DLIBMESH_DIR=${LIBMESH_INSTALL_PATH}
+RUN make -j$(nproc)
```
pshriwise: I don't see a default value for `nproc`, which means we'll use all cores available during the build step. I think a safer thing to do would be to define a default value of `1` for this variable above.

Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM ubuntu:24.04
+
+ARG NUM_THREADS=1
```
pshriwise: Name suggestion: `BUILD_JOBS`

Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+
+
+# build XDG
+RUN git clone --recurse-submodules https://github.com/xdg-org/xdg.git /XDG_TEST_SYSTEM/xdg
+WORKDIR /XDG_TEST_SYSTEM/xdg/build
+RUN cmake .. \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_C_COMPILER=mpicc \
+        -DCMAKE_CXX_COMPILER=mpicxx \
+        -DXDG_ENABLE_MOAB=ON \
```
pshriwise, line 60: Let's set an install prefix here as well to exercise that part of the process.

pshriwise (changes requested):
Looking good @magnoxemo! I think this is about ready. Just a couple small items to look at yet.

pshriwise (approved):
Thanks @magnoxemo!

Issue #234, opened 2026-05-19
Manage case where number of test failures matches expected skip return code


pshriwise:
As discussed [here]() in #215, Catch2 returns the number of test assertion failures as a return code from an executable. If this sum happens to be the code recognized by ctest as a skipped test. Thus, the failures will register as a skipped test.

A custom Catch2 test driver can solve this problem with further introspection of the test session outputs. I'll submit something to address it soon.

Waqar-ukaea:
Closed by #235

PR #235, opened 2026-05-20
Custom Catch2 test runner

pshriwise:
This PR implements a custom Catch2 runner to address #234. This is pretty niche, but I could see how we'd be lucky enough to have it confuse someone who's unaware of the issue.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -39,7 +39,7 @@ endif()
 
 foreach(test ${TEST_NAMES})
     add_executable(${test} test_main.cpp ${test}.cpp)
-    target_link_libraries(${test} PRIVATE xdg fmt::fmt Catch2::Catch2WithMain)
+    target_link_libraries(${test} PRIVATE xdg fmt::fmt Catch2::Catch2)
     if (XDG_ENABLE_LIBMESH)
         target_link_libraries(${test} PRIVATE ${LIBMESH_LINK_LIBRARIES})
     endif()
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -38,7 +38,7 @@ if (XDG_ENABLE_MOAB AND XDG_BUILD_TOOLS)
 endif()
 
 foreach(test ${TEST_NAMES})
-    add_executable(${test} ${test}.cpp)
+    add_executable(${test} test_main.cpp ${test}.cpp)
     target_link_libraries(${test} PRIVATE xdg fmt::fmt Catch2::Catch2WithMain)
     if (XDG_ENABLE_LIBMESH)
         target_link_libraries(${test} PRIVATE ${LIBMESH_LINK_LIBRARIES})
--- a/tests/test_main.cpp
+++ b/tests/test_main.cpp
@@ -1,7 +1,6 @@
 // test_main.cpp
 #define CATCH_CONFIG_RUNNER
 #include <catch2/catch_all.hpp>
-#include <iostream>
 
 // Global tracking structure to capture test summary data safely
 struct RunSummary {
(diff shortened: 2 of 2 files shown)


Waqar-ukaea (commented):
I'll save a proper review for later once the PR is ready but I had one point about linking to Catch2 with this change

tests/CMakeLists.txt
```diff
@@ -30,19 +30,15 @@ if (XDG_ENABLE_LIBMESH)
...
 endif()
 
 if (XDG_ENABLE_MOAB AND XDG_BUILD_TOOLS)
     list(APPEND TEST_NAMES test_overlap_check)
 endif()
 
 foreach(test ${TEST_NAMES})
-    add_executable(${test} ${test}.cpp)
+    add_executable(${test} test_main.cpp ${test}.cpp)
     target_link_libraries(${test} PRIVATE xdg fmt::fmt Catch2::Catch2WithMain)
```
Waqar-ukaea: I know its still in draft (and maybe you have already made this change) but since we are now providing our own "test main" should we be linking to `Catch2::Catch2` rather than `Catch2WithMain`?

pshriwise: Good call. Made that adjustment.

pshriwise:
Marking this as ready for review. Mainly made it a draft as I was wondering if it's worth the added complexity. Probably is and I'm more grappling with the fact that we have to jump through this hoop at all :-/

tests/test_main.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+// Create an EventListener to automatically capture Catch2 test run statistics
+struct TestInterceptor : Catch::EventListenerBase {
+    using EventListenerBase::EventListenerBase;
+
+    // This hook runs automatically when the entire test execution completes
+    void testRunEnded(Catch::TestRunStats const& testRunStats) override {
+        runSummary.failedAssertions = testRunStats.totals.assertions.failed;
+        runSummary.skippedTestCases = testRunStats.totals.testCases.skipped;
+        runSummary.executedTestCases = testRunStats.totals.testCases.total() - runSummary.skippedTestCases;
```
Waqar-ukaea, line 20: I guess useful in case we want to extend further but we aren't actually using these two for anything right now are we?

tests/test_main.cpp
```diff
@@ -0,0 +1,47 @@
+// test_main.cpp
+#define CATCH_CONFIG_RUNNER
+#include <catch2/catch_all.hpp>
+#include <iostream>
```
Waqar-ukaea: Not sure we actually need iostream since we aren't outputting anything

tests/test_main.cpp
```diff
@@ -0,0 +1,47 @@
...
+    int result = session.run();
+
+    // if failures were reported, but the test runner returned 4 indicating a
+    // skipped test to Ctest, increment the return code to 5 to indicate a
+    // failure to Ctest instead
+    if (runSummary.failedAssertions > 0 && result == 4) {
+        result++;
+    }
+
+    return result;
```
Waqar-ukaea: I don't think it really matters either way but any particular reason to return result+1? We could just return `1` or something right? 
```suggestion
    if (runSummary.failedAssertions > 0 && result == 4) {
        return 1;
    }

    return result;
```

Waqar-ukaea (approved):
Looks good a couple of small comments but nothing that requires a change per se. 

I'll wait for the change to link to `Catch2:Catch2` rather than `Catch2::Catch2WithMain` to be pushed before I merge though.

tests/test_main.cpp
pshriwise: Yeah that's correct. I kept them around in case we might want to print those out in the future or for use in debugging.

tests/test_main.cpp
pshriwise: I liked this because we're returning something that's closer to the true number of failures, but in practice it probably doesn't matter either way.

pshriwise: I guess in this case we can simply return a non-zero exit code if there are any failed assertions and remove the second condition.

Waqar-ukaea: Yeah that was my thoughts

pshriwise: I removed the additional conditional in this statement and we now return a consistent error code whenever an assertion fails.

PR #237, opened 2026-06-01
Hexahedral mesh support

pshriwise:
This PR implements support for hexahedral and quad elements/faces in XDG. While the changes span a lot of files, the important one lie in the mesh managers. One important thing to note is that element containment changes significantly. Where we were using barycentric coordinates for to check tet containment before, this approach now uses ~Plücker~ a dot product check against each face to determine if the query location is on the correct side of each element face. This approach assumes that the faces are linear and may need to be updated for additional face types. 

Some changes to accessing element connectivity, element vertices, and face vertices have been updated, changing from fixed size arrays to vector types to support variable number of vertex coordinates for faces/elements.

Several tests that are either equivalent to or similar to existing tet/tri models have been added to ensure functionality for these new element/face types.

One current limitation of this implementation is that all elements in a volume/faces on a surface are expected to be the same type. Support for mixed element types in a given volume will be included in later work.

**Note**: Performance testing is underway for this branch currently.


src/hexahedron_contain.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+#include "xdg/constants.h"
+#include "xdg/ray_tracing_interface.h"
+#include "xdg/ray.h"
+#include "xdg/vec3da.h"
+
+namespace xdg
+{
+
+bool plucker_hex_containment_test(const Position& point,
```
Waqar-ukaea: This method doesn't actually use plucker coords right? Would it make more sense to rename it as such? I don't actually have a suggestion for a different name however so maybe `plucker_containment` for now is fine :/

src/hexahedron_contain.cpp
Waqar-ukaea, line 1: As a general question do we want these Embree specific callbacks to all live in one file? I.e have `triangle_intersect.cpp` + `hexahedron_contain.cpp` + `quadrilateral_intersection.cpp` all sit within some file maybe just called `embree\intersection.cpp`? 

Then the more general culling logic or plucker geometry logic can be pulled out into another separate file too.

src/triangle_intersect.cpp
```diff
@@ -27,12 +32,13 @@ bool primitive_mask_cull(RTCDualRayHit* rayhit, int primID) {
...
 
   RTCSurfaceDualRay& ray = rayhit->ray;
-  RTCDualHit& hit = rayhit->hit;
 
   // if the primitive mask is set, cull if the primitive is not in the mask
   return std::find(ray.exclude_primitives->begin(), ray.exclude_primitives->end(), primID) != ray.exclude_primitives->end();
 }
 
+} // namespace
+
```
Waqar-ukaea: Why are these methods being marked by this namespace to be private to just triangle intersections? Is the same logic not useful for the other element types?

include/xdg/moab/direct_access.h
```diff
@@ -33,66 +34,140 @@ class MBDirectAccess {
...
+      fe = face_data.first_elements[block_idx];
     }
     return true;
   }
 
-  //! \brief Get the coordinates of a triangle as XDG Vertices
+  //! \brief Check that a triangle is part of the managed coordinates here
+  inline bool accessible(EntityHandle tri) {
+    return accessible(tri, SurfaceFaceType::TRI);
+  }
```
Waqar-ukaea: I can see why this wrapper is useful for preserving the previous triangle-only call path. One concern is that it now gives triangles a shorthand/default overload, while quads require explicitly passing `SurfaceFaceType::QUAD`. Is that intentional, or should callers be required to specify the face type in all cases now that we support multiple face types? 

With the wrapper we are implicitly suggesting that the default is a triangle (which given XDG/DAGMC's history is maybe fair)

Waqar-ukaea: Maybe an explicit overload for eachtype would be preferred? 


```suggestion
  //! \brief Check that a triangle is part of the managed coordinates here
  inline bool accessible_tri(EntityHandle handle) {
    return accessible(handle, SurfaceFaceType::TRI);
  }
  
  //! \brief Check that a quad is part of the managed coordinates here
  inline bool accessible_quad(EntityHandle quad) {
    return accessible(handle, SurfaceFaceType::QUAD);
  }

```

include/xdg/moab/direct_access.h
```diff
@@ -33,66 +34,140 @@ class MBDirectAccess {
...
 
     std::array<xdg::Vertex, 3> vertices;
-    vertex_data_.set_coords(i0, vertices[0]);
-    vertex_data_.set_coords(i1, vertices[1]);
-    vertex_data_.set_coords(i2, vertices[2]);
+    vertex_data_.set_coords(static_cast<int>(i0), vertices[0]);
+    vertex_data_.set_coords(static_cast<int>(i1), vertices[1]);
+    vertex_data_.set_coords(static_cast<int>(i2), vertices[2]);
     return vertices;
   }
```
Waqar-ukaea: The comment suggests that we are returning the coordinates of a tet element but this is returning coords of a triangle no?

I'm not actually certain this function is currently used anywhere but is it also worth switching to a `std::vector<Vertex>` to extend to returning quad vertices if the entity has 4 vertices (similar to what is done in other places)?

include/xdg/moab/direct_access.h
```diff
@@ -33,66 +34,140 @@ class MBDirectAccess {
...
+  //! \brief Get the connectivity of an element with explicit type
+  inline std::vector<MeshID> get_element_connectivity(const EntityHandle& element, VolumeElementType element_type) {
+    const auto& element_data = element_data_.at(element_type);
+    if (element_data.element_stride == 4) {
+      return to_mesh_ids(element_data.get_connectivity_indices<4>(element));
+    }
+    if (element_data.element_stride == 8) {
+      return to_mesh_ids(element_data.get_connectivity_indices<8>(element));
+    }
+    throw std::runtime_error("Unsupported element stride for connectivity");
```
Waqar-ukaea: Is there ever a scenario where we have the same number of vertices for two different element types? Because we call the function based on element stride rather than the enum for element type, if there is a scenario where we support two different element types which share the same number of vertices would this logic still work? 

I am less familiar with element types but are there any where this could be a concern?

include/xdg/moab/direct_access.h
```diff
@@ -33,66 +34,140 @@ class MBDirectAccess {
...
     std::array<xdg::Vertex, 4> vertices;
-    vertex_data_.set_coords(i0, vertices[0]);
-    vertex_data_.set_coords(i1, vertices[1]);
-    vertex_data_.set_coords(i2, vertices[2]);
-    vertex_data_.set_coords(i3, vertices[3]);
+    for (int i = 0; i < 4; i++) {
+      vertex_data_.set_coords(static_cast<int>(conn[i]), vertices[i]);
+    }
+    return vertices;
+  }
```
Waqar-ukaea: This seems to be similar to the same thing I highlighted for `mb_get_coords()` for faces. This appears to be a function that is used to return coords specifically for tets but the comment suggests it is for a triangle. 

Once again should a `std::vector<Vertex>` be used to pick between returning tet vertices vs hex vertices? Likewise we could just add an explicitly named extra function `get_element_coords_hex()`.

src/embree/ray_tracer.cpp
```diff
@@ -131,13 +133,22 @@ EmbreeRayTracer::register_surface(const std::shared_ptr<MeshManager>& mesh_manag
...
+    rtcSetGeometryBoundsFunction(surface_geometry, (RTCBoundsFunction)&TriangleBoundsFunc, nullptr);
+    rtcSetGeometryIntersectFunction(surface_geometry, (RTCIntersectFunctionN)&TriangleIntersectionFunc);
+    rtcSetGeometryOccludedFunction(surface_geometry, (RTCOccludedFunctionN)&TriangleOcclusionFunc);
+  } else if (surface_data->element_type == SurfaceFaceType::QUAD) {
+    rtcSetGeometryBoundsFunction(surface_geometry, (RTCBoundsFunction)&QuadBoundsFunction, nullptr);
+    rtcSetGeometryIntersectFunction(surface_geometry, (RTCIntersectFunctionN)&QuadIntersectionFunction);
+    rtcSetGeometryOccludedFunction(surface_geometry, (RTCOccludedFunctionN)&QuadOcclusionFunction);
+  } else {
+    fatal_error("Unsupported surface element type for surface {}", surface);
+  }
```
Waqar-ukaea: This is more of a general point about what we want XDG to be able to handle but this current flow inherently assumes every face in that surface in that surface is of the same type right? Do you think we would eventually want to support multiple face types for a single surface? The only example I can think of is where you might have a model with hex and wedge elements then you end up in a situation where a surface could be made up of triangles and quads right?

I _**think**_ in theory this would be possible in embree by defining multiple `rtcGeometry` objects for the same surface. Something like: 
```
RTCScene for Volume 1
  ├── RTCGeometry for Surface A triangles
  ├── RTCGeometry for Surface A quads
  ├── RTCGeometry for Surface B triangles
  ├── RTCGeometry for Surface C quads
  └── ...
```
I am definitely getting ahead of myself here but wanted to highlight this as a point regardless.

src/quadrilateral_intersection.cpp
```diff
@@ -0,0 +1,205 @@
...
+bool primitive_mask_cull(RTCDualRayHit* rayhit, int primID) {
+  if (!rayhit->ray.exclude_primitives) return false;
+
+  RTCSurfaceDualRay& ray = rayhit->ray;
+
+  // if the primitive mask is set, cull if the primitive is not in the mask
+  return std::find(ray.exclude_primitives->begin(), ray.exclude_primitives->end(), primID) != ray.exclude_primitives->end();
+}
+
+} // namespace
```
Waqar-ukaea: I think I have another comment on this inside of `triangle_intersect.cpp` about private helper functions but are these not the same for any kind of face type? If so no need to duplicate between cpp files right? I notice we have the header file `xdg\geometry\face_common.h` so this could be a good place for these methods to live?

src/quadrilateral_intersection.cpp
```diff
@@ -0,0 +1,205 @@
...
+  const PrimitiveRef& primitive_ref = user_data->prim_ref_buffer[args->primID];
+  BoundingBox bounds = mesh_manager->face_bounding_box(primitive_ref.primitive_id);
+
+  args->bounds_o->lower_x = bounds.min_x - user_data->box_bump;
+  args->bounds_o->lower_y = bounds.min_y - user_data->box_bump;
+  args->bounds_o->lower_z = bounds.min_z - user_data->box_bump;
+  args->bounds_o->upper_x = bounds.max_x + user_data->box_bump;
+  args->bounds_o->upper_y = bounds.max_y + user_data->box_bump;
+  args->bounds_o->upper_z = bounds.max_z + user_data->box_bump;
+}
```
Waqar-ukaea: Along the same lines as the previous comment in this file, the logic here will always be identical no matter the face type right? The underlying `mesh_manager->face_bounding_box()` may do something different depending on the face type but the bounds function callback logic I think will always be the same? Worth dropping the separate functions between `triangle_intersect.cpp` and `quadrilateral_intersection.cpp` in favour of some shared function `FaceBoundsFunction()` and then when setting the callback methods for `rtcGeometry` objects in `EmbreeRayTracer` we call something like:

```cpp
    rtcSetGeometryBoundsFunction(surface_geometry, (RTCBoundsFunction)&FaceBoundsFunc, nullptr);
```
irrespective of if it is triangles or quads

src/quadrilateral_intersection.cpp
```diff
@@ -0,0 +1,205 @@
...
+  RTCSurfaceDualRay& ray = rayhit->ray;
+
+  Position ray_origin = {ray.dorg[0], ray.dorg[1], ray.dorg[2]};
+  Direction ray_direction = {ray.ddir[0], ray.ddir[1], ray.ddir[2]};
+
+  std::array<Position, 3> tri0 {face_vertices[0], face_vertices[1], face_vertices[2]};
+  std::array<Position, 3> tri1 {face_vertices[0], face_vertices[2], face_vertices[3]};
+
+  auto result0 = plucker_ray_tri_intersect(tri0.data(), ray_origin, ray_direction, ray.dtfar, 0.0, false, 0);
+  auto result1 = plucker_ray_tri_intersect(tri1.data(), ray_origin, ray_direction, ray.dtfar, 0.0, false, 0);
```
Waqar-ukaea, line 46: So my understanding here is that to test for an intersection with a quad we are essentially just doing the full FP64 intersection for the two triangles that make up the quad. Is this the typical solution for handling ray-quad intersections? Or could we implement another algorithm which would be computationally cheaper than doing the full triangle intersect twice?

tests/test_ray_fire.cpp
```diff
@@ -109,4 +109,33 @@ TEMPLATE_TEST_CASE("Ray Fire on MeshMock (per-backend sections)", "[rayfire][moc
     intersection = rti->ray_fire(volume_tree, origin, direction, INFTY, HitOrientation::EXITING, &exclude_primitives);
     REQUIRE(intersection.second == ID_NONE);
   }
-}
\ No newline at end of file
+}
+
+TEST_CASE("Quad Ray Fire on MockedQuadHexMesh (Embree)")
```
Waqar-ukaea, line 114: This could be a `TEMPLATE_TEST_CASE` now and just pass in `EmbreeRayTracer` until another ray tracer backend implements quad support. But also happy to make that change later down the line when adding the quad support to the other backend too.

tools/particle_sim.h
```diff
@@ -15,7 +15,7 @@ struct SimulationData {
   uint32_t n_particles_ {100};
   uint32_t max_events_ {1000};
   bool verbose_particles_ {false};
-  bool implicit_complement_is_graveyard_ {false};
+  bool implicit_complement_is_graveyard_ {true};
```
Waqar-ukaea: Why has the default behaviour changed here?

src/libmesh/mesh_manager.cpp
```diff
@@ -80,6 +81,30 @@ void LibMeshManager::init() {
...
+    const auto& side_pair = sidepair(faces[0]);
+    SurfaceFaceType type = side_pair.face_type();
+
+    // check that all other faces are the same type
+    for (const auto face : faces) {
+      const auto& side_pair = sidepair(face);
+      if (side_pair.face_type() != type) {
+        fatal_error("Surface {} has mixed face types, which is not supported. Face ID {} has type {}, expected type {}.",
+                    surface, face, side_pair.face_type(), type);
+      }
```
Waqar-ukaea, line 105: Do we also want to do a similar initialisation time check for volume elements? I'm not sure if we have a similar error out if the mesh contains a volume with mixed types in the same way we have defined above for surfaces.

src/libmesh/mesh_manager.cpp
```diff
@@ -80,6 +81,30 @@ void LibMeshManager::init() {
   }
 
   map_id_spaces();
+
+  check_types();
+}
+
+void LibMeshManager::check_types() const {
```
Waqar-ukaea: Minor point here but `check_types()` is quite a vague name for the method. Maybe something like `check_surface_face_type_consistency()`

src/libmesh/mesh_manager.cpp
```diff
@@ -532,14 +557,74 @@ LibMeshManager::element_vertices(MeshID element) const {
...
+}
+
+VolumeElementType
+LibMeshManager::get_volume_element_type(MeshID volume) const {
+  const auto elements = get_volume_elements(volume);
+  if (elements.empty()) {
+    fatal_error("Volume {} has no elements; cannot determine element type", volume);
+  }
+
+  const auto elem_ptr = mesh()->elem_ptr(elements.front());
```
Waqar-ukaea, line 643: We are essentially doing the same thing as above in the `get_surface_face_type()` method where we get the first element and return the type of that first element. But right now because we aren't explicitly checking the consistency of all the elements in a volume is it still fair to just return the type of the first? I think we should either validate it during initialisation or least comment on the assumption here.

src/mesh_manager_interface.cpp
```diff
@@ -151,35 +150,62 @@ MeshManager::next_element(MeshID current_element,
                            const Position& r,
                            const Position& u) const
 {
-  std::array<double, 4> dists = {INFTY, INFTY, INFTY, INFTY};
-  std::array<bool, 4> hit_types;
-
   auto element_face_accessor = ElementFaceAccessor::create(this, current_element);
```
Waqar-ukaea, line 156: How would you feel about `ElementFaceAccessor` being able to return the face type? That way when switching code pathway we do so via `switch` statement with each case being the face type, rather than checking `coords.size()`.

src/mesh_manager_interface.cpp
```diff
@@ -151,35 +150,62 @@ MeshManager::next_element(MeshID current_element,
...
+                                               true,
+                                               orientation);
+      bool hit0 = result0.hit;
+      bool hit1 = result1.hit;
+      double dist0 = result0.t;
+      double dist1 = result1.t;
+      if (hit0 || hit1) {
+        hit_types[i] = true;
+        if (hit0 && hit1) dists[i] = std::min(dist0, dist1);
+        else dists[i] = hit0 ? dist0 : dist1;
```
Waqar-ukaea: Is this the exact same logic that we have in `quadrilateral_intersection.cpp`? Is it worth abstracting out the double-plucker-ray-tri intersect into a simple wrapper so that we can just call `plucker_ray_quad_intersect()` in both places? 

I'm not sure on this though, it might be better to keep things more explicit.

tests/test_mesh_connectivity.cpp
```diff
@@ -96,23 +96,23 @@ TEST_CASE("Get Surface Connectivity (MeshMock)", "[surface][unit]") {
...
+    5, 6, 7, 3,
+    5, 8, 6, 3,
+    0, 7, 1, 3,
+    0, 5, 7, 3,
+    2, 6, 8, 3,
+    2, 8, 4, 3,
+    0, 8, 5, 3,
+    0, 4, 8, 3,
+    1, 7, 6, 3,
+    1, 6, 2, 3
```
Waqar-ukaea: Should these have changed?

tests/test_mesh_internal.cpp
```diff
@@ -22,3 +22,39 @@ TEST_CASE("Test Mesh Mock")
...
+  REQUIRE_THAT(normal.y, Catch::Matchers::WithinAbs(0.0, 1e-12));
+  REQUIRE_THAT(normal.z, Catch::Matchers::WithinAbs(1.0, 1e-12));
+
+  auto bbox = mm->face_bounding_box(quad_face);
+  REQUIRE_THAT(bbox.min_x, Catch::Matchers::WithinAbs(0.0, 1e-12));
+  REQUIRE_THAT(bbox.min_y, Catch::Matchers::WithinAbs(0.0, 1e-12));
+  REQUIRE_THAT(bbox.min_z, Catch::Matchers::WithinAbs(0.0, 1e-12));
+  REQUIRE_THAT(bbox.max_x, Catch::Matchers::WithinAbs(1.0, 1e-12));
+  REQUIRE_THAT(bbox.max_y, Catch::Matchers::WithinAbs(1.0, 1e-12));
+  REQUIRE_THAT(bbox.max_z, Catch::Matchers::WithinAbs(0.0, 1e-12));
```
Waqar-ukaea: Should we also check some of these for the triangle face too?

tests/test_measure.cpp
```diff
@@ -186,3 +186,40 @@ TEMPLATE_TEST_CASE("Test Area and Volume Spherical Mesh", "[measure][sphere]",
     REQUIRE_THAT(sum_element_volumes, Catch::Matchers::WithinAbs(total_volume, 1e-6));
   }
 }
+
+TEMPLATE_TEST_CASE("TEST Quad Measurements", "[model_properties][quads]", MOAB_Interface)
```
Waqar-ukaea: Should this also be enabled for libmesh?

include/xdg/moab/direct_access.h
```diff
@@ -175,7 +249,8 @@ class MBDirectAccess {
     // ordering of element faces based on the cannonical ordering descibed here:
     // Canonical numbering systems for finite‐element codes (http://dx.doi.org/10.1002/cnm.1237)
     std::unordered_map<EntityType, std::vector<std::vector<int>>> ordering = {
-    {MBTET, {{0, 1, 3}, {1, 2, 3}, {2, 0, 3}, {0, 2, 1}}}
+      {MBTET, {{0, 1, 3}, {1, 2, 3}, {2, 0, 3}, {0, 2, 1}}},
+      {MBHEX, {{0, 1, 2, 3}, {4, 5, 6, 7}, {0, 1, 5, 4}, {1, 2, 6, 5}, {2, 3, 7, 6}, {3, 0, 4, 7}}}
```
Waqar-ukaea: Possibly one to highlight in the documentation if we aren't going to publicly expose methods to return the ordering?

include/xdg/moab/direct_access.h
```diff
@@ -33,66 +34,140 @@ class MBDirectAccess {
...
+    return element_adjacency_data_.at(element_type).get_adjacent_element(element, face_number);
+  }
+
+  inline std::vector<EntityHandle> get_element_adjacencies(const EntityHandle& element, VolumeElementType element_type) {
+    return element_adjacency_data_.at(element_type).get_element_adjacencies(element);
+  }
+
+  const std::vector<std::vector<int>>& get_face_ordering(VolumeElementType element_type) const {
+    const auto& adjacency_data = element_adjacency_data_.at(element_type);
+    return adjacency_data.ordering.at(adjacency_data.entity_type);
```
Waqar-ukaea, line 160: Broader question would it ever be useful to expose the ordering to the public API for xdg?

include/xdg/testing/mesh_mocks.h
```diff
@@ -262,6 +327,16 @@ class MeshMock : public MeshManager {
...
 
+  // 0 0
+  // 1 1
+  // 2 2
+  // 8 3
+  // 3 4
+  // 4 5
+  // 6 6
+  // 5 7
+  // 7 8
```
Waqar-ukaea: Should these be here and if so, what are they?

include/xdg/testing/mesh_mocks.h
```diff
@@ -303,20 +378,171 @@ class MeshMock : public MeshManager {
...
-  const MeshMock* mesh_manager_;
+  const MockedTriTetMesh* mesh_manager_;
   std::array<std::array<int, 3>, 4> tet_connectivity_;
 };
+
+namespace xdg {
+
+// A lightweight mesh for testing face representation (tri + quad).
+class MockedQuadHexMesh : public MeshManager {
+public:
```
Waqar-ukaea: Currently this class is just testing Tris + Quads? I don't see any data for hex elements. Will this be coming in a subsequent PR?

include/xdg/testing/mesh_mocks.h
```diff
@@ -303,20 +378,171 @@ class MeshMock : public MeshManager {
...
+
+  // Connectivity is CCW when viewed from +Z, yielding +Z normals (outward for a volume on -Z side).
+  const std::unordered_map<MeshID, std::vector<MeshID>> face_connectivity_ {
+    {0, {0, 1, 2, 3}}, // quad (+Z normal)
+    {1, {4, 5, 6}},    // tri (+Z normal)
+  };
+
+  const std::unordered_map<MeshID, std::vector<MeshID>> surface_faces_ {
+    {0, {0}},
+    {1, {1}},
```
Waqar-ukaea: In it's current state is this the mock mesh data structured like this? 
```
volume 0
  surface 0: one quad face
  surface 1: one triangular face
```
So the mock is representing a mixed face type mesh?

tests/test_libmesh.cpp
Waqar-ukaea: Worth moving to `util.h` in case other tests want to make use of it?

src/hexahedron_contain.cpp
Waqar-ukaea: We have a number of tests for the `plucker_tet_containment_test()` but none that I can currently see for the `plucker_hex_containment_test()`. 

Could we `TEMPLATE_TEST_CASE` the tests in `test_tet_containment.cpp` to also test for hexes? I haven't looked into the test too much so maybe not directly like that. But maybe renaming the test file to `test_element_containment.cpp` and it could include test cases for tet and hex containments?

Waqar-ukaea: Likewise can `test_tet_intersection.cpp` receive similar treatment to also test for a hex intersection?

tests/test_libmesh.cpp
```diff
@@ -276,6 +285,60 @@ TEST_CASE("Test Point Location Jezebel")
...
+  std::shared_ptr<XDG> xdg = XDG::create(MeshLibrary::LIBMESH);
+  xdg->mesh_manager()->mesh_library();
+  REQUIRE(xdg->mesh_manager()->mesh_library() == MeshLibrary::LIBMESH);
+  const auto& mesh_manager = xdg->mesh_manager();
+  mesh_manager->load_file("jezebel-quads.exo");
+  mesh_manager->init();
+  xdg->prepare_raytracer();
+
+  Position inside {0.0, 0.0, 0.0};
+  REQUIRE(xdg->find_element(inside) != ID_NONE);
```
Waqar-ukaea: Could we instead check that `find_element(inside)` returns the expected element ID here rather than only checking that it does return non `ID_NONE`?

tests/test_libmesh.cpp
```diff
@@ -496,4 +559,71 @@ TEST_CASE("Multiblock sidesets")
   auto tracks = xdg->segments(volume, start, end);
 
   assert(tracks.size() > 0);
-}
\ No newline at end of file
+}
+
+TEMPLATE_TEST_CASE("TEST libMesh Raytrace Quads", "[moab][faces][quads]",
```
Waqar-ukaea: This has the wrong test label:
```suggestion
TEMPLATE_TEST_CASE("TEST libMesh Raytrace Quads", "[libMesh][faces][quads]",
```

tests/test_libmesh.cpp
```diff
@@ -496,4 +559,71 @@ TEST_CASE("Multiblock sidesets")
...
+    // fire rays in random directions and verify that we get a hit at
+    // approximately the expected distance of 6.3849 (the radius of the sphere)
+    for (int i = 0; i < 10; ++i) {
+      dir = rand_dir();
+      auto hit = xdg->ray_fire(volume, origin, dir);
+      REQUIRE_THAT(hit.first, Catch::Matchers::WithinAbs(6.3849, 1e-2));
+      REQUIRE(hit.second != ID_NONE);
+    }
+  }
+}
```
Waqar-ukaea: As far as I can tell, this test is identical to the one inside of `test_moab.cpp` for raytracing against quads. This _could_ instead live inside of `test_xdg_interface.cpp` as a `TEMPLATE_PRODUCT_TEST_CASE`. I think this is maybe a wider question about where certain tests should live between `test_libmesh` and `test_moab` and any changes could probably come in a separate PR.

tests/test_libmesh.cpp
```diff
@@ -496,4 +559,71 @@ TEST_CASE("Multiblock sidesets")
...
+
+    SimulationData sim_data;
+    sim_data.xdg_ = xdg;
+    sim_data.n_particles_ = 1000;
+    sim_data.mfp_ = 0.5;
+    sim_data.verbose_particles_ = false;
+    sim_data.implicit_complement_is_graveyard_ = true;
+    transport_particles(sim_data);
+  }
+}
```
Waqar-ukaea, line 686: We aren't actually making any assertions here on correctness of the transport right? So I guess this is moreso just testing whether or not the code crashes for each file?

tests/test_moab.cpp
```diff
@@ -183,110 +188,292 @@ TEMPLATE_TEST_CASE("TEST MOAB Find Element Method", "[moab][elements]",
...
+  static constexpr int32_t num_elements = 0;
+  static constexpr int32_t num_faces = 2474;
+  static constexpr int32_t num_vertices = 1580;
+};
+struct CylBrickQuadsH5M {
+  static constexpr std::string_view filename = "cyl-brick-quads.h5m";
+  static constexpr int32_t num_elements = 0;
+  static constexpr int32_t num_faces = 20478;
+  static constexpr int32_t num_vertices = 196869;
+};
```
Waqar-ukaea, line 285: I like the idea of these structs with encoded metadata for the test files. Do you think it would be worth pulling them out into some central location so that they can be used from within other test files? 

We could also consider storing the model stem (for example `jezebel`) and deriving variants such as `.h5m`, `-quads.h5m` or `-tet.h5m` where those files exist. Though this idea may be a little overkill for not much gain.

tests/test_moab.cpp
```diff
@@ -3,18 +3,23 @@
 #include <cmath>
 #include <memory>
 #include <numeric>
+#include <random>
+#include <string_view>
 
 // testing includes
 #include <catch2/catch_template_test_macros.hpp>
 #include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
```
Waqar-ukaea: Do we still need this header? I don't see any instances of generators being used.

src/moab/mesh_manager.cpp
```diff
@@ -408,32 +459,82 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
...
+MOABMeshManager::get_surface_face_type(MeshID surface) const
 {
-  moab::EntityHandle surf_handle = this->surface_id_map_.at(surface);
+  auto faces = this->_surface_faces(surface);
+
+  if (faces.all_of_type(moab::MBTRI)) {
+    return SurfaceFaceType::TRI;
+  }
+  if (faces.all_of_type(moab::MBQUAD)) {
+    return SurfaceFaceType::QUAD;
```
Waqar-ukaea: Are these `all_of_type()` calls expensive? And do they scale with the number of faces? As far as I understand by the time we reach this point in the code we already have checks to ensure that every surface stored is not of mixed face types at initiliasation and I wonder whether if relying on this `all_of_type()` check is worth it here.

Waqar-ukaea: Notably when it comes to volume elements I can see we are just checking the first element and returning that type. 

I suppose the current setup will be valuable once we reach a point of having mixed face types in a single surface though.

src/moab/mesh_manager.cpp
```diff
@@ -408,32 +459,82 @@ MOABMeshManager::get_volume_surfaces(MeshID volume) const
...
-  moab::EntityType type = moab::MBTRI; // TODO: hardcodeed to tri for now
+  moab::EntityHandle element_handle;
+  auto rval = this->moab_interface()->handle_from_id(moab::MBTET, elements.front(), element_handle);
+  if (rval == moab::MB_SUCCESS) {
+    return VolumeElementType::TET;
+  }
+
+  rval = this->moab_interface()->handle_from_id(moab::MBHEX, elements.front(), element_handle);
+  if (rval == moab::MB_SUCCESS) {
+    return VolumeElementType::HEX;
```
Waqar-ukaea, line 500: Somewhat related to my comment on questioning the use of `all_of_type()` is there a reason that method is not used here for volume elements? Does it only work with faces? I think it makes sense to be consistent in how we decide which type to return for a volume or surface

include/xdg/testing/mesh_mocks.h
```diff
@@ -171,8 +233,15 @@ class MeshMock : public MeshManager {
...
-    return SurfaceElementType::TRI; // hardcoded to Tri for this mock
+  virtual SurfaceFaceType get_surface_face_type(MeshID surface) const override {
+    return SurfaceFaceType::TRI; // hardcoded to Tri for this mock
+  }
+
+  virtual VolumeElementType get_volume_element_type(MeshID volume) const override {
+    if (get_volume_elements(volume).empty()) {
+      fatal_error("MockedTriTetMesh volume {} has no elements; cannot determine element type", volume);
+    }
+    return VolumeElementType::TET;
```
Waqar-ukaea: Just to stay consistent with the `get_surface_face_type()`
```suggestion
    return VolumeElementType::TET; // hardcoded to Tet for this mock
```

include/xdg/testing/mesh_mocks.h
```diff
@@ -303,20 +378,171 @@ class MeshMock : public MeshManager {
...
+    return surface_sense_map_.at(surface);
+  }
+
+  Sense surface_sense(MeshID surface, MeshID volume) const override {
+    auto senses = surface_senses(surface);
+    if (volume == senses.first) return Sense::FORWARD;
+    if (volume == senses.second) return Sense::REVERSE;
+    fatal_error("Volume {} is not a parent of surface {}", volume, surface);
+    return Sense::UNSET;
+  }
```
Waqar-ukaea, line 461: Should the TetMeshMock also have a similar method?

include/xdg/testing/mesh_mocks.h
```diff
@@ -303,20 +378,171 @@ class MeshMock : public MeshManager {
...
+    return count;
+  }
+  int num_surface_faces(MeshID surface) const override {
+    return static_cast<int>(surface_faces_.at(surface).size());
+  }
+  int num_vertices() const override { return static_cast<int>(vertices_.size()); }
+
+  std::vector<MeshID> get_volume_elements(MeshID /*volume*/) const override { return {}; }
+  std::vector<MeshID> get_volume_connectivity(MeshID /*volume*/) const override { return {}; }
+  std::vector<Vertex> get_volume_vertices(MeshID /*volume*/) const override { return {}; }
```
Waqar-ukaea: Should these also return a warning to suggest that the mock has no volume elements in the same way `get_volume_element_type()` does?

include/xdg/moab/direct_access.h
```diff
@@ -195,29 +270,29 @@ class MBDirectAccess {
...
+      if (!entity_range.all_of_type(entity_type)) { throw std::runtime_error("Not all elements match the requested type"); }
 
       moab::Range::iterator entity_it = entity_range.begin();
       while(entity_it != entity_range.end()) {
         // set connectivity pointer, element stride and the number of elements
         EntityHandle* conntmp;
         int n_elements;
         rval = mbi->connect_iterate(entity_it, entity_range.end(), conntmp, element_stride, n_elements);
-        MB_CHK_SET_ERR_CONT(rval, "Failed to get direct access to triangle elements");
+        MB_CHK_SET_ERR_CONT(rval, "Failed to get direct access to elements");
```
Waqar-ukaea, line 274: Just to clarify `elements` here is referring to both face elements and volume elements? As the method (and surrounding struct for `ConnectivityData` can represent both?

Waqar-ukaea (commented):
Really nice stuff here @pshriwise and I'm sure @nuclearkevin will be excited to get stuck in with adding further element types following this PR! 

I do have quite a few comments but most of these are checking consistency or small suggestions for some minor code cleanup. Pretty happy with the major changes here across the board.

Waqar-ukaea:
> Where we were using barycentric coordinates for to check tet containment before, this approach now uses Plücker coordinates to check that the query location is on the correct side of each element face. This approach assumes that the faces are linear and may need to be updated for additional face types.

This comment seems to suggest that the containment logic for tet_containment checks has changed but as far as I can tell that file hasn't been touched in this PR?

src/hexahedron_contain.cpp
pshriwise: I don't have a strong preference here. Separate files seems a little more friendly to someone looking at the codebase for the first time, but I'm happy either way. If we do make that move I'd like to do it in a separate PR so it doesn't complicate the diff here.

src/hexahedron_contain.cpp
pshriwise: Mmmm yes. This is true. I think my memory was failing me here. I'll drop the "plucker" from the name here. Plucker coordinates don't really apply b/c they require an orientation for the query to compute.

src/quadrilateral_intersection.cpp
pshriwise: Yeah I think I did this originally b/c I wasn't sure if they would turn out to be the exact same at first but never went back and closed the loop on it. Thanks for noticing!

include/xdg/moab/direct_access.h
pshriwise: Yeah, agreed that this isn't the most extensible design, but I think the bigger question might be whether or not we keep them given that they aren't being called anywhere. But if we do keep them I'd probably just opt to keep the one version and have it require a face type.

include/xdg/moab/direct_access.h
pshriwise: My best guess is that the comment got out of place after a rebase. I don't see a call to it either and if it's dead code... 👋

include/xdg/moab/direct_access.h
pshriwise: Some can, yes, but I don't foresee us supporting some of the more "exotic" element types where this could be an issue (e.g. a trigonal bipyramid).

Using the `VolumeElementType` is more straightforward and much more clear so I'll switch to that 👍🏻

include/xdg/moab/direct_access.h
pshriwise: Another place where I think comments got shuffled during a rebase. I don't see a reason to keep this either given that it's only called to get element volumes (which were limited to tets) and upon digging deeper it made me realize that we should probably abstract the entire `element_volume` method to the interface class so I've done that.

src/embree/ray_tracer.cpp
pshriwise: I think we discussed this on our latest call, but yes for this initial implementation of new element/face type this limitation exists -- many of the error messages reflect this. I of course meant to put that in the PR but forgot.

src/quadrilateral_intersection.cpp
pshriwise: I'm open to additional discussion on this but this is a pretty standard approach and will suffice for now I think. I'd flag this for future research/discussion if you're alright with that.

tests/test_ray_fire.cpp
pshriwise: Yeah I think I'd rather leave the additional code out until it's functional.

tools/particle_sim.h
pshriwise: Oh that was a quick change to test some files but didn't flip it back. Thanks for catching it!

include/xdg/testing/mesh_mocks.h
pshriwise: These are related to the changes in the mesh connectivity in the test you commented on above. It was a temporary mapping I jotted down but it shouldn't remain.

src/hexahedron_contain.cpp
Waqar-ukaea: Hmm, thats a good point. I'm happy to stick with the separate files then but I do think they can probably be moved into the `/embree` directory because as far as I an tell they are entirely embree specific. And agreed, that can happen in a separate PR

tests/test_mesh_connectivity.cpp
pshriwise: Long story short, no these should not have changed and I made a mistake in my latest rebase by keeping some of the concrete implementations of methods like `get_volume_vertices()` that were abstracted to a higher level in one of your recent PRs.

There are a few other abstracted methods like this as well that I'll clean up too.

src/libmesh/mesh_manager.cpp
pshriwise: Yeah that's fair. Let me know what you think of the new name.

src/moab/mesh_manager.cpp
pshriwise: It's a good note. It may become irrelevant once we start looking at mixed element types as this method doesn't really lend itself to support for mixed types in general. 

The concern about performance here is a good one. I think I'll change it to match that of the `get_volume_element_type` for the time being.

pshriwise: I'll also update the default return here to something more meaningful (even though it's behind an error call that will shut things down before we return).

tests/test_mesh_internal.cpp
pshriwise: I think the triangle in that mesh was me playing around with mixed element types and then realizing that a better incremental step was to work with consistent element types for given mesh regions. 

I've updated the this mesh mock class to be hex/quad only and added an extra element to allow for more testing.

include/xdg/moab/direct_access.h
pshriwise: Yep. I'll make a note of it.

include/xdg/testing/mesh_mocks.h
pshriwise: Cleaned this up generally and made it hex/quad only as the name indicates.

pshriwise: There's a reply elsewhere with more detail.

include/xdg/moab/direct_access.h
pshriwise: I don't see any harm in it, but given that issue was already present based on existing data was already here I'm going to call that decision out of scope for this PR. Happy to discuss further though!

include/xdg/testing/mesh_mocks.h
pshriwise: https://github.com/xdg-org/xdg/pull/237#discussion_r3416557524

include/xdg/testing/mesh_mocks.h
pshriwise: https://github.com/xdg-org/xdg/pull/237#discussion_r3416557524

tests/test_libmesh.cpp
pshriwise: This isn't being used anywhere. In light of that, I'm going to remove it.

include/xdg/moab/direct_access.h
Waqar-ukaea: Sounds good

src/libmesh/mesh_manager.cpp
Waqar-ukaea: I'm happy with that

tests/test_mesh_internal.cpp
Waqar-ukaea: Okay yeah that makes more sense

src/quadrilateral_intersection.cpp
pshriwise: Yeah this makes sense. I'll consolidate these into one in a new source file. Placing all of these files containing the Embree-specific callbacks makes sense to me as well.

src/libmesh/mesh_manager.cpp
pshriwise: This has been updated so that the two methods match. We may need to switch back to the "`all_of_type`" approach later for mixed elements but this guarantees efficiency for now.

src/mesh_manager_interface.cpp
pshriwise: I think I'd prefer to keep it explicit. I'll take another look at this and see if I can clean it up a bit though. It's a little hard to follow currently.

pshriwise: I did find some places we can make this more clear and, along the way, some places where we can optimize things a bit. If you'll allow I'd like to save that for an immediate follow-on PR though.

src/mesh_manager_interface.cpp
pshriwise: I think yes. Somewhat tied to the below comment on repeated code related to face intersections below. 

https://github.com/xdg-org/xdg/pull/237#discussion_r3389979662

You okay if I include it in a follow-on PR?

tests/test_measure.cpp
pshriwise: Done!

src/moab/mesh_manager.cpp
pshriwise: Hopefully resolved in the other related conversation for surface face types.

include/xdg/testing/mesh_mocks.h
pshriwise: I think it does? Though they don't look exactly the same.

include/xdg/testing/mesh_mocks.h
pshriwise: I went ahead and implemented these so I think we're good.

include/xdg/moab/direct_access.h
pshriwise: Yep yep that's correct.

src/hexahedron_contain.cpp
pshriwise: Yep! Implemented. I did find some redundant testing in `test_tet_intersection` (it was actually testing containment) and ended up removing the file.

tests/test_libmesh.cpp
pshriwise: Correct. Just a light robustness test.

tests/test_moab.cpp
pshriwise: >I like the idea of these structs with encoded metadata for the test files. Do you think it would be worth pulling them out into some central location so that they can be used from within other test files?

I like this idea, but I realized that by putting this data in the XDG repo it's at risk of becoming stale relative to the actual test files. I'm going to create a header in the test files repo we can include linstead.

> We could also consider storing the model stem (for example jezebel) and deriving variants such as .h5m, -quads.h5m or -tet.h5m where those files exist. Though this idea may be a little overkill for not much gain.

I'm going to vote overkill on this one :)

src/libmesh/mesh_manager.cpp
pshriwise: Thanks for asking about this! It revealed an issue with testing ID spaces for libMesh meshes (which was a pain to fix). But it's been resolved now.

pshriwise:
Okay @Waqar-ukaea I think this is ready for another look. I made a few issues related to follow-on PRs we discussed throughout:

- [ ] Addition of a model_data.h header to go in the XDG files repo https://github.com/xdg-org/xdg_test_files/issues/18
- [ ] Document use of canonical element ordering based on Tautges paper #242 
- [ ] Consolidate Embree source files into the embree directory #243 
- [ ] ElementFaceAccessor returning face type #244

include/xdg/geometry/face_common.h
```diff
@@ -2,6 +2,7 @@
 #define XDG_GEOMETRY_TRIANGLE_INTERSECT_H
 
 #include <array>
+#include "xdg/shared_enums.h"
```
Waqar-ukaea: Seems like this header isn't needed? Possibly a hold out from before the addition of the `intersect_common.h`?

include/xdg/libmesh/mesh_manager.h
```diff
@@ -61,7 +61,18 @@ class LibMeshManager : public MeshManager {
...
+  static VolumeElementType _elem_xdg_type(const libMesh::Elem *elem_ptr) {
+    switch (elem_ptr->type()) {
+    case libMesh::TET4:
+      return VolumeElementType::TET;
+    case libMesh::HEX8:
+      return VolumeElementType::HEX;
+    default:
+      return VolumeElementType::UNSUPPORTED;
+    }
+  }
```
Waqar-ukaea, line 75: I like this as a private helper, curious if a similar thing would also be useful for faces?

include/xdg/moab/direct_access.h
Waqar-ukaea: Did you want to make this change to remove these methods since they aren't being called/switching to the version which explicitly requests type in this PR or a future one?

include/xdg/moab/direct_access.h
```diff
@@ -326,23 +303,30 @@ class MBDirectAccess {
...
+          auto conn = data.get_connectivity_indices<3>(face);
+          std::array<EntityHandle, 3> verts = {conn[0], conn[1], conn[2]};
+
+          Range adjacent_elements;
+          rval = mbi->get_adjacencies(verts.data(), verts.size(), 3, true, adjacent_elements);
+          MB_CHK_SET_ERR_CONT(rval, "Failed to get MOAB boundary face adjacencies");
+
+          // sanity check: an element face should have at most two adjacent elements
+          if (adjacent_elements.size() > 2) {
+            throw std::runtime_error("Found more than two elements adjacent to a face");
```
Waqar-ukaea: Super minor point here but I noticed that within this file we seem to throw `std::runtime_error` and I'm just curious why we don't rely on `fatal_error()` like we usually do. Are these runtime errors exceptions caught anywhere and handled or are they acting as unrecoverable exits out of the code?

src/bounding_functions.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+  const PrimitiveRef& primitive_ref = user_data->prim_ref_buffer[args->primID];
+  BoundingBox bounds = mesh_manager->face_bounding_box(primitive_ref.primitive_id);
+
+  args->bounds_o->lower_x = bounds.min_x - user_data->box_bump;
+  args->bounds_o->lower_y = bounds.min_y - user_data->box_bump;
+  args->bounds_o->lower_z = bounds.min_z - user_data->box_bump;
+  args->bounds_o->upper_x = bounds.max_x + user_data->box_bump;
+  args->bounds_o->upper_y = bounds.max_y + user_data->box_bump;
+  args->bounds_o->upper_z = bounds.max_z + user_data->box_bump;
```
Waqar-ukaea, line 41: I'm curious about the difference in bump handling here between face `SurfaceFaceBounds` and `VolumeElementBounds`. 

With surface faces my understanding is that we attach the bump to the `user_data` object inside `EmbreeRayTracer::create_surface_tree()` by calling `bump = bounding_box_bump(mesh_manager, volume_id)`. And that function itself is described in `ray_tracing_interface.cpp` as:
```cpp
const double RayTracer::bounding_box_bump(const std::shared_ptr<MeshManager> mesh_manager, MeshID volume_id)
{
  auto volume_bounding_box = mesh_manager->volume_bounding_box(volume_id);
  return std::max(volume_bounding_box.dilation(), numerical_precision_);
}
```
Whereas for the volume elements we are calculating the bump directly inside the `VolumeElementBoundsFunc()` and directly calling `BoundingBox::dilation()` instead of `std::max(dilation(), numerical_precision_)`. So we are being inconsistent between the two types of bounding box right? Is this intentional?

src/intersect_common.cpp
Waqar-ukaea, line 1: One for a future PR but these could be made GPU friendly in the future and we could re-use the same functions directly within the traversal algorithm for both CPU and GPU RT backends.

src/mesh_manager_interface.cpp
Waqar-ukaea: Yep sure

src/mesh_manager_interface.cpp
```diff
@@ -292,6 +278,21 @@ MeshManager::surface_bounding_box(MeshID surface) const
   return bb;
 }
 
+double MeshManager::element_volume(MeshID element) const
+{
+  auto vertices = this->element_vertices(element);
```
Waqar-ukaea: This line isn't used because we later face the face vertices directly from the element_face_accessor in the loop.
```suggestion
```

src/quadrilateral_intersection.cpp
Waqar-ukaea: Agreed

tests/test_tet_intersection.cpp
Waqar-ukaea, line 1: Presumably this was deleted because the same test coverage was covered more clearly inside of `test_test_containment.cpp`?

src/embree/ray_tracer.cpp
```diff
@@ -133,17 +134,17 @@ EmbreeRayTracer::register_surface(const std::shared_ptr<MeshManager>& mesh_manag
...
-  if (surface_data->element_type == SurfaceFaceType::TRI) {
-    rtcSetGeometryBoundsFunction(surface_geometry, (RTCBoundsFunction)&TriangleBoundsFunc, nullptr);
+  if (surface_data->face_type == SurfaceFaceType::TRI) {
+    rtcSetGeometryBoundsFunction(surface_geometry, (RTCBoundsFunction)&SurfaceFaceBoundsFunc, nullptr);
     rtcSetGeometryIntersectFunction(surface_geometry, (RTCIntersectFunctionN)&TriangleIntersectionFunc);
     rtcSetGeometryOccludedFunction(surface_geometry, (RTCOccludedFunctionN)&TriangleOcclusionFunc);
-  } else if (surface_data->element_type == SurfaceFaceType::QUAD) {
-    rtcSetGeometryBoundsFunction(surface_geometry, (RTCBoundsFunction)&QuadBoundsFunction, nullptr);
+  } else if (surface_data->face_type == SurfaceFaceType::QUAD) {
+    rtcSetGeometryBoundsFunction(surface_geometry, (RTCBoundsFunction)&SurfaceFaceBoundsFunc, nullptr);
```
Waqar-ukaea: Both branches now call `rtcSetGeometryBoundsFunction(surface_geometry, (RTCBoundsFunction)&SurfaceFaceBoundsFunc, nullptr);` so we can pull this line out to before the `if-else` block rather than calling it in both.

tests/test_mesh_internal.cpp
Waqar-ukaea, line 27: Would we want a similar Face Representation test for the `TriTestMeshMock`? Maybe some of the logic is covered in other tests but I think at least some stuff like the testing each surface is of the correct type or returns the number of vertices might be missing.

tests/test_tet_containment.cpp
Waqar-ukaea: Super super minor point here but should this be renamed to match the hex test case? 
```cpp
TEST_CASE("Tetrahedron Containment Unit Test")
```

Waqar-ukaea (commented):
A few more comments @pshriwise but looks like we're pretty much there

include/xdg/libmesh/mesh_manager.h
pshriwise: For faces it's already wrapped into a method under the `SidePair` object. Would be nice if there were a convenient place to mirror this, but give the presence of `SidePair::face_type` I'm not sure where we'd call it.

include/xdg/moab/direct_access.h
pshriwise: I think one snuck back in in in my latest rebase. Removing now. Thanks!

include/xdg/moab/direct_access.h
pshriwise: Oh I think that probably is there from before I implemented `fatal_error`. We should always fail through that interface instead. Updating.

src/bounding_functions.cpp
pshriwise: > Is this intentional?

Yes, very. For surface trees, we need to account for the error based on the max distance a ray can travel before hitting a surface (the conservative estimate being the max chord length of the volume's bounding box.

For element trees where we're performing point queries there's no reason to expand the box more than a small amount based on the local element's original bounding box as there's no additional dilation to add based on ray length. We only need to expand the box enough to ensure that if the query point is on a vertex/face of the element in double precision that we won't skip over it in single precision during traversal.

src/intersect_common.cpp
pshriwise: They're pretty simple functions but I'm all for code consolidation!

src/mesh_manager_interface.cpp
pshriwise: Good catch!

tests/test_tet_intersection.cpp
pshriwise: Correct. I thought I made a comment along those lines. Apologies if I didn't.

tests/test_mesh_internal.cpp
pshriwise: Sure I can add a test there for parity.

pshriwise:
Okay I think I hit all of your latest comments. Thank you for the detailed review! I know it's maybe our biggest diff to date.

src/bounding_functions.cpp
Waqar-ukaea: Ah okay i see. Cool that makes sense

include/xdg/moab/direct_access.h
Waqar-ukaea: Looks like there is still one `std::runtime_error()` here on line 334

pshriwise:
Update:

I found that the current algorithm used in hex containment wasn't sufficient for robust element walking of hexes. Due to small but significant numerical differences (not just floating point) the points of quad faces aren't exactly planar in many cases. This requires us to both a) check both triangles when considering point containment for hexes/performing ray distance queries on quad faces and b) use a consistent diagonal across the face so that queries aren't dependent on the connectivity ordering of the face depending on the element we're querying that face from (i.e. the winding of the face may be correct in many orderings, but the diagonal used to separate the face into triangles will change). I've introduced a vertex sort based on lexicographical ordering of the vertex coordinates for now though I'd like to change to another method (probably vertex IDs) for a consistent ordering that doesn't require a lot of floating point comparison later on.

I caught this after trying some basic transport with the `walk-elements` tool and to check problems like this more quickly in the future I've added some tests that walk elements using the same code as that tool.

Waqar-ukaea:
> use a consistent diagonal across the face so that queries aren't dependent on the connectivity ordering of the face depending on the element we're querying that face from

Interesting, I see. Presumably this would also affect the robustness of ray-quad intersections for surface tracking as well right? If the diagonal is inconsistent depending on which volume you are firing the ray against?

> I caught this after trying some basic transport with the walk-elements tool and to check problems like this more quickly in the future I've added some tests that walk elements using the same code as that tool.

Perfect, I was just about to suggest this! If there is also a robustness issue with surface-tracking for quads maybe some tests to stress that edge case would be useful too?

pshriwise:
> Interesting, I see. Presumably this would also affect the robustness of ray-quad intersections for surface tracking as well right? If the diagonal is inconsistent depending on which volume you are firing the ray against?
> 

Somewhat, yes. If the model contains only hexes (only possible in MOAB meshes at the moment), then the connectivity is always consistent and it's just a matter of us returning reversed normal for volumes with the appropriate sense.

In the context of a hex mesh and, for example, firing a ray to determine re-entry into the mesh from the implicit complement we do need to ensure that the quad is treated the same for both the ray fire operation to determine the entry location _and_ the subsequent hex operations for walking elements/determining tally segments. Therefore we peform the same ordering operation in the quad intersection callbacks as well.

> Perfect, I was just about to suggest this! If there is also a robustness issue with surface-tracking for quads maybe some tests to stress that edge case would be useful too?

Yeah, good idea. I'll add it to the current set of particle tracking tests that reside in the mesh backend test files.

src/mesh_manager_interface.cpp
```diff
@@ -1,15 +1,37 @@
...
+    const Vertex& b_max = lower(b0, b1) ? b1 : b0;
+
+    if (lower(a_min, b_min)) return true;
+    if (lower(b_min, a_min)) return false;
+    return lower(a_max, b_max);
+  };
+
+  return pair_less(coords[0], coords[2], coords[1], coords[3]);
+}
+
```
Waqar-ukaea: This is identical to the method in `face_common.h` for determining the quad diagonal right? Is it intentional that we have two methods for this?

pshriwise: Relying on a single implementation in `face_common.h` now. Thanks!

PR #194, opened 2026-01-30
[PR4] Ray-benchmark tool to test throughput of batch query API added in #195

Waqar-ukaea:
This PR adds a new tool/miniapp that I have been using to benchmark the batch query API pure ray throughput on various GPUs. 

This PR adds `tools/ray_benchmark/` which includes:
- `ray_benchmark.cpp` - The main application 
- `ray_benchmark.h` - Header file containing some helper functions including definition of the callback method which "mocks" an external application filling XDG's internal ray buffers
- `ray_benchmark_deviceCode.slang` - Slang compute shader for filling XDG's internal ray buffer on device
- `ray_benchmark_shared.h` - Shared code between host and device
- `ray_benchmark_driver.py` - A python script used to drive multiple runs of the `ray_benchmark` miniapp for more consistent results measurement. 

Copied from #178:

### Benchmark parameters
| model        | Volume | No. of Elements | No. of Rays | Location | No. of Runs
|--------------------|--------|-----------------|-------------|-------------|----------------|
| simple_tokamak| 2      | 280K            | 50M         | (180,250,-27) | 100 |

A render of the simple_tokamak model [1] used in these preliminary benchmarks along with a depiction (significantly smaller number of rays plotted) of the rays launched is shown in the image below, along with the volume queried against highlghted in blue:
ray_benchmark_tokamak_setup
### Ray tracer performance (trace-only)  
Baseline = **Embree (CPU), 2× Intel® Xeon® Platinum 8480+ (Sapphire Rapids) × 112 threads**

Times and throughput averaged over the 100 runs.

Ray tracer performance (trace-only).

Baseline = Embree (CPU), 2x Intel Xeon Platinum 8480+ (Sapphire Rapids), 112 threads.

| Ray Tracer backend | Hardware (Threads / Device) | Trace Time (s) | Throughput (ray/s) | Speedup vs 2x8480+ (112-thread) | Peak FP32/FP64 (TFLOPS) + RT cores | Node % |
|---|---|---:|---:|---:|---|---:|
| Embree | 13th Gen Intel Core i7-13850HX x **28** threads | 0.749128 | 1.06791e+08 | ~0.25x | N/A | Not HPC |
| Embree | 1x Intel Xeon Platinum 8480+ (Sapphire Rapids) x **56** threads | 0.367881 | 2.17462e+08 | ~0.51x | N/A | 50% |
[...]

PR #238, opened 2026-06-03
Ray benchmark miniapp

Waqar-ukaea:
This PR adds a simple miniapp to compare the total ray throughput between backends within XDG. 

I previously had a PR up for something similar in #194 centred around GPRT. The difference here in this new PR is that the benchmark miniapp is being added independent of any GPU backend (GPRT or cuBQL) and only works with embree along with some minor changes/cleanup to get it to a state where it is much easier to extend to also work with cuBQL. 

One immediate question is whether or not an explicit check for correctness is required in this benchmark. Initially I didn't want to do anything that checks for correct rays as it would slow down the benchmark but I wonder if that would be useful.

### DIFF
--- a/tools/ray_benchmark.cpp
+++ b/tools/ray_benchmark.cpp
@@ -19,6 +19,8 @@
 #include "xdg/xdg.h"
 
 #include "ray_benchmark.h"
+#include <fmt/ranges.h>
+
 
 using namespace xdg;
 
@@ -126,9 +128,7 @@ int main(int argc, char** argv)
   mesh_manager->init();
 
   if (args.get<bool>("--list")) {
-    for (auto mesh_volume : mesh_manager->volumes()) {
-      std::cout << mesh_volume << std::endl;
-    }
+    std::cout << "[" << fmt::format("{}", fmt::join(mesh_manager->volumes(), ", ")) << "]\n";
     return 0;
   }
 
--- a/tools/ray_benchmark.cpp
+++ b/tools/ray_benchmark.cpp
@@ -154,7 +154,23 @@ int main(int argc, char** argv)
 
   xdg->prepare_volume_for_raytracing(volume);
   xdg->ray_tracing_interface()->init();
+
+  const bool origin_in_volume = xdg->point_in_volume(volume, origin);
+  if (!origin_in_volume) {
+    const std::string origin_label = source_radius > 0.0 ? "source center" : "ray origin";
+    warning(fmt::format("The {} ({}, {}, {}) is not inside volume {}. Results may be skewed by fewer intersections with the BVH.",
+                        origin_label, origin.x, origin.y, origin.z, volume));
+  } else {
+    const double nearest_surface_distance = xdg->closest_distance(volume, origin);
+    if (source_radius > nearest_surface_distance) {
+      warning(fmt::format("The source radius ({}) is larger than the nearest surface distance ({}) from the source center to volume {}. "
+                          "Some sampled source points may be outside the volume.",
+                          source_radius, nearest_surface_distance, volume));
+    }
+  }
+
   setup_timer.stop();
+
   if (rt_lib == RTLibrary::EMBREE) {
     rt_label += " (" + std::to_string(XDGConfig::config().n_threads())
              + " CPU threads)";
@@ -162,39 +178,38 @@ int main(int argc, char** argv)
 
   const auto num_faces = mesh_manager->num_volume_faces(volume);
 
-  if (rt_lib == RTLibrary::EMBREE) {
-    // Generate random rays from source
-    generation_timer.start();
-    std::vector<Position> origins(num_rays);
-    std::vector<Direction> directions(num_rays);
 
-    #pragma omp parallel for schedule(runtime)
-    for (std::size_t i = 0; i < num_rays; ++i) {
-      std::uint32_t state = seed ^ static_cast<std::uint32_t>(i);
-      auto sample = tools::benchmark::random_spherical_source(origin.x,
-                                                              origin.y,
-                                                              origin.z,
-                                                              state,
-                                                              source_radius);
-      origins[i] = Position(sample.position[0],
-                            sample.position[1],
-                            sample.position[2]);
-      directions[i] = Direction(sample.direction[0],
-                                sample.direction[1],
-                                sample.direction[2]);
-    }
-    generation_timer.stop();
+  // Generate random rays from source
+  generation_timer.start();
+  std::vector<Position> origins(num_rays);
+  std::vector<Direction> directions(num_rays);
 
-    // Trace rays
-    trace_timer.start();
-
-    #pragma omp parallel for schedule(runtime)
-    for (std::size_t i = 0; i < num_rays; ++i) {
-      (void) xdg->ray_fire(volume, origins[i], directions[i]);
-    }
-
-    trace_timer.stop();
+  #pragma omp parallel for schedule(runtime)
+  for (std::size_t i = 0; i < num_rays; ++i) {
+    std::uint32_t state = seed ^ static_cast<std::uint32_t>(i);
+    auto sample = tools::benchmark::random_spherical_source(origin.x,
+                                                            origin.y,
+                                                            origin.z,
+                                                            state,
+                                                            source_radius);
+    origins[i] = Position(sample.position[0],
+                          sample.position[1],
+                          sample.position[2]);
+    directions[i] = Direction(sample.direction[0],
+                              sample.direction[1],
+                              sample.direction[2]);
   }
+  generation_timer.stop();
+
+  // Trace rays
+  trace_timer.start();
+
+  #pragma omp parallel for schedule(runtime)
+  for (std::size_t i = 0; i < num_rays; ++i) {
+    (void) xdg->ray_fire(volume, origins[i], directions[i]);
+  }
+
+  trace_timer.stop();
 
   const double generation_time = generation_timer.elapsed();
   const double trace_time = trace_timer.elapsed();
@@ -258,7 +273,7 @@ int main(int argc, char** argv)
     std::cout << fmt::format("{}\n", fmt::join(csv_columns, ","));
     std::cout << fmt::format("{}\n", fmt::join(csv_values, ","));
   } else {
-    std::cout << "XDG ray benchmark\n";
+    std::cout << "\nXDG ray benchmark\n";
     std::cout << "----------------------------------------\n";
     std::cout << "Model                  : " << model_name << "\n";
     std::cout << "Mesh library           : " << mesh_str << "\n";
--- a/tools/ray_benchmark.cpp
+++ b/tools/ray_benchmark.cpp
@@ -1,5 +1,6 @@
 #include <cstdint>
 #include <cstdlib>
+#include <filesystem>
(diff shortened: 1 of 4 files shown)


tools/ray_benchmark.cpp
```diff
@@ -0,0 +1,213 @@
...
+    .help("Number of rays to cast for the benchmark")
+    .scan<'u', std::uint32_t>();
+
+  args.add_argument("-s", "--seed")
+    .default_value<std::uint32_t>(12345)
+    .help("Seed for random ray generation")
+    .scan<'u', std::uint32_t>();
+
+  args.add_argument("-o", "-p", "--origin", "--position")
+    .default_value(std::vector<double>{0.0, 0.0, 0.0})
```
pshriwise: What if we have this default to the center of the volume's bounding box? That could either be a new option or could be the default value of the ray origin?

tools/ray_benchmark.cpp
```diff
@@ -0,0 +1,213 @@
...
+  // XDG setup and ray tracer initialisation
+  setup_timer.start();
+  std::shared_ptr<XDG> xdg = XDG::create(mesh_lib, rt_lib);
+  const auto& mesh_manager = xdg->mesh_manager();
+  mesh_manager->load_file(args.get<std::string>("filename"));
+  mesh_manager->init();
+
+  if (args.get<bool>("--list")) {
+    for (auto mesh_volume : mesh_manager->volumes()) {
+      std::cout << mesh_volume << std::endl;
```
pshriwise: We'll want some kind of delimiter and spacing here.

tools/ray_benchmark.cpp
```diff
@@ -0,0 +1,213 @@
...
+    return 0;
+  }
+
+  xdg->prepare_volume_for_raytracing(volume);
+  xdg->ray_tracing_interface()->init();
+  setup_timer.stop();
+
+  if (rt_lib == RTLibrary::EMBREE) {
+    #ifdef XDG_OPENMP
+    rt_label += " (" + std::to_string(omp_get_max_threads()) + " CPU threads)";
```
pshriwise: Let's use the `xdg::config` object here instead of a direct OpenMP call.

tools/ray_benchmark.h
```diff
@@ -0,0 +1,71 @@
...
+}
+
+inline SourceSample random_spherical_source(double origin_x,
+                                            double origin_y,
+                                            double origin_z,
+                                            std::uint32_t state,
+                                            double source_radius)
+{
+  SourceSample sample;
+  random_unit_dir_lcg(state, sample.direction);
```
pshriwise, line 49: There's a method on the `Direction` object that does the same. We should probably select one approach for generating random directions and stick with it.

tools/ray_benchmark.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include "xdg/vec3da.h"
+#include "xdg/xdg.h"
+
+#include "ray_benchmark.h"
+
+using namespace xdg;
+
+int main(int argc, char** argv)
+{
+  argparse::ArgumentParser args("XDG Ray Tracing throughput benchmarking tool",
```
pshriwise:
```suggestion
  argparse::ArgumentParser args("XDG Raytracing throughput benchmarking tool",
```

tools/ray_benchmark.cpp
```diff
@@ -0,0 +1,213 @@
...
+  const double trace_only_rps = trace_time > 0.0
+    ? static_cast<double>(num_rays) / trace_time
+    : 0.0;
+  const double end_to_end_rps = end_to_end_time > 0.0
+    ? static_cast<double>(num_rays) / end_to_end_time
+    : 0.0;
+
+  wall_timer.stop();
+
+  std::cout << "Random ray generation time   = "
```
pshriwise: I know we report it above, but let's report the initialization time here as well. 

I'll suggest we also support output of a format that's parsable by tools like NumPy/Pandas. Most natural one that comes to mind would be CSV.

pshriwise (changes requested):

tools/ray_benchmark.cpp
```diff
@@ -0,0 +1,213 @@
...
+      directions[i] = Direction(sample.direction[0],
+                                sample.direction[1],
+                                sample.direction[2]);
+    }
+    generation_timer.stop();
+
+    // Trace rays
+    trace_timer.start();
+
+    #pragma omp parallel for schedule(static)
```
pshriwise: I'll suggest we stick to `schedule(runtime)` here so we can adjust the schedule settings using environment variables without recompiling the code. This can have a significant effect on some machines.

pshriwise (changes requested):
Thanks @waqar-ukaea! It will be nice to have a single tool to do some of our benchmarking work.

tools/ray_benchmark.cpp
Waqar-ukaea: I like that idea. I decided to go with the default being the center of the entire model's bounding box but also added an additional flag that sets the origin to the volume we are querying against bounding box if no explicit origin is provided.

tools/ray_benchmark.cpp
Waqar-ukaea: I've gone for using fmt here

tools/ray_benchmark.cpp
Waqar-ukaea: How would i go about doing this correctly? Because when I call `XDGConfig::config().n_threads()` from within the benchmark miniapp it returns `1`. 

I notice that inside of the setup for config we call:
```cpp
void XDGConfig::initialize() {
  // if threads aren't manually specified,
  // set using OpenMP if available, otherwise 1
  // this is here entirely to make sure that
  // libMesh respects the OpenMP settings of the host application
  if (n_threads() == -1) {
  #ifdef XDG_HAVE_OPENMP
    set_n_threads(omp_get_num_threads());
  #else
    set_n_threads(1);
  #endif
  }
  ```
But should this call `omp_get_max_threads()` instead of `omp_get_num_threads()`. Making that change does mean that in `ray_benchmark.cpp` we do report the correct number of threads (for my system 28).

Comment (Waqar-ukaea) [reply tools/ray_benchmark.cpp]: [fc46415](https://github.com/xdg-org/xdg/pull/238/commits/fc464158d36727a598b22ce4048a2bfe8ae9c53e) is the commit where I try to. It's just a problem in reporting the number of threads though as far as I can tell. The actual omp pragma region does seem to use the number of threads provided because if I change `OMP_NUM_THREADS` as an environment variable it does speedup or slowdown relative to the number of threads specified.

File: tools/ray_benchmark.h (hunk 4 above)
Comment (Waqar-ukaea) [reply tools/ray_benchmark.h:49]: I explained it poorly when we had our last catchup meeting but the reason I didn't opt for using the method on the `Direction` object was because I wasn't sure it was very device friendly. It certainly isn't slang friendly, I remember that from when I was working with GPRT. My understanding is that the LCG RNG is good for device side loops since its random per ray so each ray can derive its own state from the seed and ray id/loop counter. But the method on Direction is reliant on a host only library (i think?) for `drand48()` and is more global state random so in a parallel loop will be less consistent based on thread timing/scheduling. 

So I would prefer to stick with the method I have defined in here but perhaps can be applied to the method in `vec3da.h`?

File: tools/ray_benchmark.cpp (hunk 6 above)
Comment (Waqar-ukaea) [reply tools/ray_benchmark.cpp]: I've gone ahead and added a flag to switch between csv output and a more human readable one

**Comment by Waqar-ukaea** [comment]
@pshriwise I have gone ahead and made the changes you suggested and fleshed out some of them with some extra functionality. The only thing is with switching to using `XDGConfig` the tool now reports the incorrect number of threads when running

File: tools/ray_benchmark.cpp (hunk 2 above)
Comment (pshriwise) [reply tools/ray_benchmark.cpp]: Even better

File: tools/ray_benchmark.cpp (hunk 3 above)
Comment (pshriwise) [reply tools/ray_benchmark.cpp]: Ah yeah you are correct. Let's update that call here.

File: tools/ray_benchmark.h (hunk 4 above)
Comment (pshriwise) [reply tools/ray_benchmark.h:49]: Yeah if we're going to be generating ray state on device then I think this makes sense. LCG is suitable for the reasons you stated, though those properties are true of many PRNGs.

Comment (Waqar-ukaea) [reply tools/ray_benchmark.h:49]: Do you have a preferred PRNG use? Or happy with LCG in this instance?

Comment (pshriwise) [reply tools/ray_benchmark.h:49]: No no, LCG is great! Maybe overkill for this purpose but that's the right side to err on :) Sorry, didn't mean to imply we should change it.

File: tools/ray_benchmark.cpp (hunk 1 above)
Comment (pshriwise) [reply tools/ray_benchmark.cpp]: Follow-up thought here. Maybe we should emit a warning if the ray origin (either default or user-specified) ends up being outside the volume as it's likely to skew the benchmark results due to fewer intersections with the BVH.

Comment (pshriwise) [reply tools/ray_benchmark.cpp]: Not sure how to square this with the source radius option though. I'd say it's overkill to be checking every point we sample..

**Review by pshriwise — APPROVED** [review]
Just one final suggestion, but regardless left but regardless I think this is good to go!

File: tools/ray_benchmark.cpp [hunk 8]
```diff
@@ -0,0 +1,285 @@
...
+  xdg->ray_tracing_interface()->init();
+  setup_timer.stop();
+  if (rt_lib == RTLibrary::EMBREE) {
+    rt_label += " (" + std::to_string(XDGConfig::config().n_threads())
+             + " CPU threads)";
+  }
+
+  const auto num_faces = mesh_manager->num_volume_faces(volume);
+
+  if (rt_lib == RTLibrary::EMBREE) {
```
pshriwise: We've already ensured that this condition is true above so until we implement this for additional libraries do we need the if block here?

Waqar-ukaea: Hmm, I guess not.

tools/ray_benchmark.cpp
Waqar-ukaea: Agreed on checking single origins. I had actually had an idea for handling both of these:
```cpp
if (!xdg->point_in_volume(volume, origin) {
   warning("Origin outside");
}
else { 
   double nearest_surf_distance = xdg->closest_distance(volume, origin);
   if (source_radius > nearest_surf_distance) { 
      warning("Source radius larger than nearest surface distance from origin. Some points may lie outside the volume and lead to a reduced number of intersections");
   }
}
```
Do you think that would be enough of warning?

Waqar-ukaea: Could also be worth having the miniapp report the number of intersections too?

pshriwise: > Agreed on checking single origins. I had actually had an idea for handling both of these:

I like it! Nice idea.

> Could also be worth having the miniapp report the number of intersections too?

Shouldn't alter the timing too much. Definitely worth reporting missed rays I'd say.

Waqar-ukaea:
@pshriwise I think we are good on this one now. I may put up an additional PR that adds a python script to drive multiple runs of the benchmark too once I am at the point where I want to start doing proper comparison runs with cuBQL etc. 

I'll also close #194 in favour of this one now too.

pshriwise (approved):
Looks good! Thank you @Waqar-ukaea!

PR #248, opened 2026-07-21
First MFEM backend PR

sean-baccas:
This PR introduces MFEM as a mesh manager.

The MFEM mesh manager class was inspired heavily by the one made for libmesh. Some of the functionality was missed, most notably the ability to detect implicit boundaries within a mesh. It can handle explicitly labelled internal boundaries just fine.

The above text is a placeholder. I'll add some more substance soon.


.github/workflows/ci.yml
```diff
@@ -10,13 +10,14 @@ on:
...
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         libmesh: [true, false]
         moab: [true, false]
+        mfem: [true, false]
         exclude:
           - moab: false
             libmesh: false
```
Waqar-ukaea, line 23: We need to add an exclude for mfem too: 
```
```suggestion
matrix:
libmesh: [true, false]
moab: [true, false]
        mfem: [true, false]
exclude:
- moab: false
libmesh: false
mfem: false
```
Currently we are excluding the case `moab=false, libmesh=false, mfem=true` but this change should ensure that job runs too

tests/CMakeLists.txt
```diff
@@ -31,6 +31,10 @@ if (XDG_ENABLE_LIBMESH)
     list(APPEND TEST_NAMES test_libmesh)
 endif()
 
+if (XDG_ENABLE_MFEM)
+    list(APPEND TEST_NAMES test_mfem)
+endif()
+
 if (XDG_ENABLE_MOAB AND XDG_ENABLE_LIBMESH)
     list(APPEND TEST_NAMES test_mesh_library_cross_check)
 endif()
```
Waqar-ukaea: This currently doesn't resolve scenarios where mfem is disabled, since the cross checks include `xdg_mfem` instatiation. I'm not sure what the best way to fix this is now that we have three backends rather than two. 

The only way I can see to do it is at runtime check which backends are enabled and add them to a list of XDG instances to create. I implemented something similar for the ray_tracing_cross_check_tests with cuBQL here - https://github.com/Waqar-ukaea/xdg/blob/b857939d316fdb9ea005eedb49af9bac3e04356f/tests/test_ray_tracer_cross_check.cpp.

tests/test_mesh_library_cross_check.cpp
```diff
@@ -130,4 +131,92 @@ TEST_CASE("Test MOAB-libMesh Cross-Check Tallies -- Simple Cubes, Tet Mesh")
...
+  xdg_libmesh->mesh_manager()->load_file("jezebel.exo");
+  xdg_libmesh->mesh_manager()->init();
+  xdg_libmesh->mesh_manager()->parse_metadata();
+  xdg_libmesh->prepare_raytracer();
+
+  auto xdg_mfem = XDG::create(MeshLibrary::MFEM);
+  xdg_mfem->mesh_manager()->load_file("jezebel.exo");
+  xdg_mfem->mesh_manager()->init();
+  xdg_mfem->mesh_manager()->parse_metadata();
+  xdg_mfem->prepare_raytracer();
```
Waqar-ukaea, line 154: I have another comment talking about this. But my idea is to do something like - https://github.com/Waqar-ukaea/xdg/blob/b857939d316fdb9ea005eedb49af9bac3e04356f/tests/test_ray_tracer_cross_check.cpp

Waqar-ukaea (commented):
Minor points on CI/testing

tests/CMakeLists.txt
sean-baccas: I am going to put a hotfix in here for now, and set it so that the cross checks only run when all three backends are enabled. I'll fix that one before taking it out of draft.

CMakeLists.txt
```diff
@@ -369,6 +392,10 @@ if (XDG_ENABLE_MOAB)
   target_link_libraries(xdg PRIVATE MOAB)
 endif()
 
+if (XDG_ENABLE_MFEM)
+  target_link_libraries(xdg PUBLIC mfem)
+endif()
+
```
Waqar-ukaea: Can this be linked privately to match what we do for libmesh and moab?

tests/CMakeLists.txt
```diff
@@ -31,7 +31,11 @@ if (XDG_ENABLE_LIBMESH)
     list(APPEND TEST_NAMES test_libmesh)
 endif()
 
-if (XDG_ENABLE_MOAB AND XDG_ENABLE_LIBMESH)
+if (XDG_ENABLE_MFEM)
+    list(APPEND TEST_NAMES test_mfem)
+endif()
+
+if (XDG_ENABLE_MOAB AND XDG_ENABLE_LIBMESH AND XDG_ENABLE_MFEM)
```
Waqar-ukaea: We do need to come up with a better solution here. Having the test only be compiled when all three mesh managers are enabled is not ideal.

.github/workflows/ci.yml
```diff
@@ -117,6 +133,16 @@ jobs:
...
 
+      # - name: mfem Cache
+      #   if: ${{ matrix.mfem }}
+      #   id: mfem-cache
+      #   uses: actions/cache@v3
+      #   env:
+      #     cache-name: mfem-cache
+      #   with:
+      #     path: ~/mfem
+      #     key: mfem-${{ runner.os }}-${{ env.cache-name }}-${{ env.MFEM_SHA }}
```
Waqar-ukaea: Should these remain commented out?

.github/workflows/ci.yml
```diff
@@ -141,6 +167,23 @@ jobs:
...
+          # first let's get netcdf
+          sudo apt-get update
+          sudo apt-get install -y libnetcdf-dev libhdf5-dev
+          cd ~
+          cd mfem
+          git checkout v4.9
+          mkdir build
+          cd build
+          # manually entering the hdf5 dir since cmake can't find it on its own
+          cmake .. -DMFEM_USE_HYPRE=OFF -DMFEM_USE_NETCDF=ON -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_BUILD_TYPE=Debug -DHDF5_DIR=/usr/lib/x86_64-linux-gnu/hdf5/serial -DBUILD_SHARED_LIBS=ON  -DCMAKE_POSITION_INDEPENDENT_CODE=ON
```
Waqar-ukaea: Two changes here I suggest. Set build type to be `Release` and have the install prefix match the same style as the other installed libraries. 


```suggestion
          cmake .. -DMFEM_USE_HYPRE=OFF -DMFEM_USE_NETCDF=ON -DCMAKE_INSTALL_PREFIX=$HOME/MFEM -DCMAKE_BUILD_TYPE=Release -DHDF5_DIR=/usr/lib/x86_64-linux-gnu/hdf5/serial -DBUILD_SHARED_LIBS=ON  -DCMAKE_POSITION_INDEPENDENT_CODE=ON
```

Not sure if these changes will need to be propagated elsewhere in the CI so test before applying.

As for the HDF5 needing an explicit path, do we have any idea what is causing this? Problem with the way MFEM is searching for hdf5?

.github/workflows/ci.yml
```diff
@@ -141,6 +167,23 @@ jobs:
           make -j4
           sudo make install
 
+      - if: ${{ matrix.mfem && steps.mfem-cache.outputs.cache-hit != 'true' }}
+        name: Build mfem
+        shell: bash
+        run: |
+          # first let's get netcdf
+          sudo apt-get update
+          sudo apt-get install -y libnetcdf-dev libhdf5-dev
```
Waqar-ukaea: Let's move these to apt-dependencies step near the top of the file

.github/workflows/ci.yml
```diff
@@ -157,7 +200,13 @@ jobs:
             fi
             CMAKE_PREFIX_PATH="${CMAKE_PREFIX_PATH}$HOME/MOAB"
           fi
-          cmake .. -DCMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH" -DCMAKE_INSTALL_PREFIX=$HOME/opt -DXDG_ENABLE_MOAB=${{ matrix.moab && 'ON' || 'OFF' }} -DXDG_ENABLE_LIBMESH=${{ matrix.libmesh && 'ON' || 'OFF' }}
+          if [ "${{ matrix.mfem }}" = 'true' ]; then
+            if [ -n "$CMAKE_PREFIX_PATH" ]; then
+              CMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH;"
+            fi
+            CMAKE_PREFIX_PATH="${CMAKE_PREFIX_PATH}$HOME/mfem/install"
+          fi
```
Waqar-ukaea: With the earlier suggestion applied this should be changed to:

```suggestion
          if [ "${{ matrix.mfem }}" = 'true' ]; then
            if [ -n "$CMAKE_PREFIX_PATH" ]; then
              CMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH;"
            fi
            CMAKE_PREFIX_PATH="${CMAKE_PREFIX_PATH}$HOME/MFEM"
          fi          
```

include/xdg/bbox.h
```diff
@@ -33,6 +33,15 @@ bool operator ==(const BoundingBox& other) {
...
 
+void update(const double* v) {
+  min_x = std::min(min_x, v[0]);
+  min_y = std::min(min_y, v[1]);
+  min_z = std::min(min_z, v[2]);
+  max_x = std::max(max_x, v[0]);
+  max_y = std::max(max_y, v[1]);
+  max_z = std::max(max_z, v[2]);
+}
+
```
Waqar-ukaea:
```suggestion
```
I'm not sure this is used anywhere so happy to remove?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,437 @@
...
+      fatal_error("Unsupported geom");
+  }
+}
+
+// Same problem as above. Volume is supposed to be a block_id, and this
+// function is interpreting it as an element index. We should have a LOT more
+// elements than blocks, so it's safe, but wrong. When dealing with a mixed
+// mesh, it will fail tests
+VolumeElementType MfemMeshManager::get_volume_element_type(MeshID volume) const {
+  mfem::Geometry::Type geom = mesh_->GetElementBaseGeometry(volume);
```
Waqar-ukaea: I am understanding the comment suggesting that this mfem method is expecting to be passed an element ID rather than a volume ID? 

If so can we do something like 
```suggestion
  auto element = volume_to_element_map.at(volume)[0];
  mfem::Geometry::Type geom = mesh_->GetElementBaseGeometry(element);
```

Doesn't fix the mixed mesh issue but at least we are passing an ID of the correct dimension?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,437 @@
...
+    }
+  }
+}
+
+// This is not quite correct. mfem does support mixed meshes.
+// The intention of the caller is that the argument (surface) corresponds
+// to the sideset. So we need to find a typical element from the sideset
+// that is marked by the argument surface. For now, this will do
+SurfaceFaceType MfemMeshManager::get_surface_face_type(MeshID surface) const {
+  mfem::Geometry::Type geom = mesh_->GetFaceGeometry(surface);
```
Waqar-ukaea: Same as below comment?

src/config.cpp
Waqar-ukaea: I think this needs to be extended to include MFEM

```suggestion
bool XDGConfig::mesh_manager_enabled(MeshLibrary mesh_lib) const {
  #ifdef XDG_ENABLE_MOAB
  if (mesh_lib == MeshLibrary::MOAB) return true;
  #endif
  #ifdef XDG_ENABLE_LIBMESH
  if (mesh_lib == MeshLibrary::LIBMESH) return true;
  #endif
  #ifdef XDG_ENABLE_MFEM
  if (mesh_lib == MeshLibrary::MFEM) return true;
  #endif
  return false;
}
```

src/config.cpp
```diff
@@ -44,33 +44,7 @@ void XDGConfig::initialize() {
...
-  // register for cleanup at program exit
-
-  // libMesh expects to be able to clean some static objects up at exit if they
-  // aren't present, so we register this cleanup function to be called at exit
-  // before those objects are deleted by libMesh. Otherwise, a double-free can
-  // occur.
-  std::atexit(cleanup_libmesh_at_exit);
-#endif
-
-initialized_ = true;
```
Waqar-ukaea: I assume previous behaviour was causing this to always be called no matter if a `LibMeshManager` was ever constructed? Hence the reason for moving this down into `libmesh_init()`?

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  // count the number of faces with the attribute "volume"
+  int num_volume_faces(MeshID volume) const override {
+    fatal_error("MfemMeshManager::num_volume_faces() not implemented yet");
+  }
+
+  int num_surface_faces(MeshID surface) const override {
+    fatal_error("MfemMeshManager::num_surface_faces() not implemented yet");
+  }
```
Waqar-ukaea: If we can figure out how to implement this method, it will go a long way. When we create construct the GPU ray tracers (GPRT and cuBQL) we rely on methods at the abstract `MeshManager` layer to return flattened array of vertices and connectivity indices between those vertices. This backend specific method lets us ensure the correct bounds are applied when constructing the flattened arrays.

Can we do something like `sideset_to_element_map.at(surface).size()`

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+    return output;
+  }
+
+  virtual std::vector<MeshID> element_connectivity(MeshID element) const override {
+    fatal_error("MfemMeshManager::element_connectivity() not implemented yet");
+  }
+
+  virtual std::vector<MeshID> face_connectivity(MeshID face) const override {
+    fatal_error("MfemMeshManager::face_connectivity() not implemented yet");
+  }
```
Waqar-ukaea: This method is the other one required to build up the flattened arrays needed to construct GPU ray tracers with XDG but upon further reflection I think we already essentially get this information from a different method anyway.

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+  int num_surface_faces(MeshID surface) const override {
+    fatal_error("MfemMeshManager::num_surface_faces() not implemented yet");
+  }
+
+  // get all of the elements in this volume
+  virtual std::vector<MeshID> get_volume_elements(MeshID volume) const override;
+
+  virtual std::vector<MeshID> get_surface_faces(MeshID surface) const override;
+
+  virtual std::vector<Vertex> element_vertices(MeshID element) const override;
```
Waqar-ukaea: Add a space here to separate the subsequent method for readability
```suggestion
  virtual std::vector<Vertex> element_vertices(MeshID element) const override;
  
```

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+      case 3: return num_volumes();
+      case 2: return num_surfaces();
+      default: return 0;
+    }
+  }
+
+  // I think we need to count the number of elements with the attribute "volume"
+  int num_volume_elements(MeshID volume) const override {
+    fatal_error("MfemMeshManager::num_volume_elements() not implemented yet");
+  }
```
Waqar-ukaea: Would `volume_to_element_map.at(volume).size()` work?

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+
+  std::pair<MeshID, MeshID> surface_senses(MeshID surface) const override;
+
+  // Seems like it's only used to create the implicit complement
+  MeshID create_volume() override;
+
+  void add_surface_to_volume(MeshID volume, MeshID surface, Sense sense, bool overwrite=false) override;
+
+  // This will be largely a copy of the libmesh version
+  void determine_surface_senses();
```
Waqar-ukaea: This is being pretty pedantic as technically a user shouldn't necessarily be interfacing directly with `MFEMMeshManager` but I think this method should probably be marked private to the class (and perhaps the same can be said for the libmesh equivalent). @pshriwise thoughts on that?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+      surface_senses_[surface_id] = {
+        mesh_->GetAttribute(face_el_tx->Elem1No), mesh_->GetAttribute(face_el_tx->Elem2No)
+      };
+    }
+
+    // We have a surface on a true boundary. The second element is simply
+    // the implcit complement
+    else {
+      surface_senses_[surface_id] = {volume, ID_NONE};
+    }
```
Waqar-ukaea: The implicit complement does have an explicit ID and as such the sense returned for it should too. `ID_NONE` is reserved for invalid IDs.

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+    fatal_error("MfemMeshManager::face_connectivity() not implemented yet");
+  }
+
+  // Topology
+
+  std::vector<MeshID> get_volume_surfaces(MeshID volume) const override;
+
+  std::pair<MeshID, MeshID> surface_senses(MeshID surface) const override;
+
+  // Seems like it's only used to create the implicit complement
```
Waqar-ukaea: Yep, we should probably have a think about that more broadly. Currently the API is named as though it could support the creation of more volumes but you're right as far as i'm aware that method is used one in implicit complement creation. @pshriwise

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+  // recogonises. volumes_ will be amended to include the
+  // implicit complement as well...
+  std::set<int> attributes_;
+
+  int num_interior_faces_;
+  int num_boundary_faces_;
+
+  //! Mapping of surfaces to the volumes on either side. Volumes are ordered
+  //! based on their sense with respect to the surface triangles. We reuase
+  //! whichever ordering mfem decides on when the mesh is constructed.
```
Waqar-ukaea: Typo
```suggestion
  //! Mapping of surfaces to the volumes on either side. Volumes are ordered
  //! based on their sense with respect to the surface triangles. We reuse
  //! whichever ordering mfem decides on when the mesh is constructed.
```

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+  // The table works wonders for this
+  virtual MeshID adjacent_element(MeshID element, int face) const override;
+
+  virtual MeshID get_boundary_face_element(MeshID face) const override {
+    fatal_error("MfemMeshManager::get_boundary_face_element() not implemented yet");
+  }
+
+  virtual Sense surface_sense(MeshID surface, MeshID volume) const override {
+    fatal_error("MfemMeshManager::surface_sense() not implemented yet");
+  }
```
Waqar-ukaea: I think we can make an argument here to have this live inside the base `MeshManager` class instead since in both the MOAB and the libMesh implementations it is just going to call the respective `surface_senses()` method for the particular mesh manager and then either assign `Sense::FORWARD` or `Sense::REVERSE`. @pshriwise any thoughts on that?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+
+    // simply return the empty vector if this is the implicit
+    // complement
+    return output;
+  }
+
+  // gather all the element IDs that have this attribute
+  // this method is absolutely criminal. Could be done at the start
+  // when we run over all the elements anyway...
```
Waqar-ukaea: lol at this method being criminal

But also do we not already do this on line 35 where we insert the elements into `volume_to_element_map_`? 

In which case can this method simply return the vector returned by `volume_to_element_map_.at_volume)`?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+  for (int i=0; i<mesh_->GetNE(); i++) {
+    if ( mesh_->GetAttribute(i) == volume ) output.push_back(i);
+  }
```
Waqar-ukaea: > this method is absolutely criminal

lol

But also do we not already do this on line 35 where we insert the elements into `volume_to_element_map_`? 

In which case can this method simply return the vector returned by `volume_to_element_map_.at_volume)`?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+namespace xdg {
+void MfemMeshManager::load_file(const std::string &filepath) {
+  mesh_ = std::make_unique<mfem::Mesh>(filepath.c_str(), 1, 1);
+}
+
+void MfemMeshManager::init() {
+  // Ensure the mesh is 3-dimensional
+  if (mesh_->Dimension() != 3) {
+    fatal_error("Mesh must be 3-dimensional");
+  }
```
Waqar-ukaea: So with MOAB at least we have the ability to support meshes which contain only surface elements. I.e volumes containing surfaces and surfaces containing triangles/quads but no interior volumetric elements. Would this check still allow for that functionality? Or would this require that a mesh contains tets/hexes?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+  
+  // set the volumes/attributes...
+  // This set should have one entry per volume/attribute type
+  for (int i=0; i<mesh_->attributes.Size(); i++) {
+    attributes_.insert( mesh_->attributes[i] );
+  }
+
+  // Create a set for each volume attribute. Gather the IDs of all the
+  // interior elements with this characteristic
+  // TODO: This won't work with ParMesh
```
Waqar-ukaea: What is ParMesh?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+    sideset_to_element_map_[sideset].insert(i);
+
+    // We also want to count the number of sidesets that each volume has.
+    // So, while we are looping over each boundary element, we look at
+    // which sideset it's a member of. We then look at
+    // its immediate neighbour on the interior of the mesh. We query
+    // this neighbour for which volume it's a member of, and register
+    // this sideset as a member of that volume.
+    // We want each sideset to be a member of exactly one volume, but
+    // that's probably too much to ask.
```
Waqar-ukaea: I'm not sure this comment is required. In XDG we want to treat shared surfaces as belonging to two parent volumes. So in that case I don't think we do want each sideset to be a member of exactly one volume

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+  }
+
+  // let's see what we got
+  std::cout << "\n\nPrinting the boundaries we discovered!!!\n";
+  auto iter = implicit_bdr.begin();
+  for (; iter!=implicit_bdr.end(); iter++) {
+    std::cout << "detected bdr between vol " << iter->first << " and " << iter->second << "\n";
+  }
```
Waqar-ukaea: This is commented out. Should it be removed?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+  // walk the surface senses and return the surfaces that have this volume
+  // as an entry
+  std::vector<MeshID> surfaces;
+  for (const auto& [surface, senses] : surface_senses_) {
+    if (senses.first == volume || senses.second == volume) {
+      surfaces.push_back(surface);
+    }
+  }
+  return surfaces;
+}
```
Waqar-ukaea: A wider question about default behaviour but this seems to me something that can live in the `MeshManager` base class. If there is an MFEM specific way that can provide a more efficient override then that can be defined here. @pshriwise ?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+  // ask the mesh for the vertices of this element
+  mesh_->GetElementVertices(element, index_array);
+
+  std::vector<Vertex> output(index_array.Size());
+
+  for (int i=0; i<index_array.Size(); i++) {
+    const double* vertices = mesh_->GetVertex( index_array[i] );
+
+    for (int d=0; d<mesh_->SpaceDimension(); d++) output[i][d] = vertices[d];
+  }
```
Waqar-ukaea: I'm a little confused on what this loop is for? What is `SpaceDimension`? Should this even be a loop at all? My reading on the logic here is that we are copying the vertex xyz components, is that correct?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+
+  return output;
+}
+
+// I've written this extra function because the mesh manager needs to support one
+// continuous list of elements for boundary and interior. So at some point we need
+// to map them together.
+// Update: not sure that's true. moab mesh manager reports the same number of elements
+// on the jezebel as mesh->GetNE()
```
Waqar-ukaea: You're right it isn't actually needed for us to support a continuous list of both interior and boundary. 

Rather we consider them two different entity types. Both are referenced via a MeshID within XDG. But we essentially have this separation which requires two sets of methods for each: 

- Volumetric elements --> Interior elements
- Surface faces --> Boundary elements

My preference is to refer to them as volume primitives and surface primitives because that's the nomenclature the ray tracing side of things works with. 

**So this method can probably go tbh. I'm not sure. Maybe it still has some use @pshriwise ?**

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+    for (int d=0; d<mesh_->SpaceDimension(); d++) output[i][d] = vertices[d];
+  }
+
+  return output;
+}
+
+MeshID MfemMeshManager::adjacent_element(MeshID element, int face) const {
+  // Would be nice if we had the face element accessor still available
+  // TODO: do something novel if we are already on the boundary
+  mfem::Array<int> faces, ori; // don't care about ori
```
Waqar-ukaea: `ori` is orientation?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+    surface_senses_[surface] = {volume, senses.second};
+  }
+
+  else {
+    if (!overwrite && senses.second != ID_NONE) {
+      fatal_error("Surface already has a reverse sense");
+    }
+      surface_senses_[surface] = {senses.first, volume};
+  }
+}
```
Waqar-ukaea: Maybe another set of candidates to be pulled into the `MeshManager` base class @pshriwise ?

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+  // list of all the attributes we've seen. Let's copy the contents of
+  // attributes_ into volumes_, so the base class has access to the list
+  // of volume IDs
+  std::copy(attributes_.begin(), attributes_.end(), std::back_inserter(volumes_));
+
+  // ditto for surfaces
+  std::copy(sideset_ids.begin(), sideset_ids.end(), std::back_inserter(surfaces_));
+
+  // set these two attributes related to interior/boundary faces
+  num_interior_faces_ = mesh_->GetNumFaces();
```
Waqar-ukaea: Does `mfem::Mesh::GetNumFaces()` explicitly exclude boundary faces? Because if not then it would be counting all faces including boundary faces right?

CMakeLists.txt
Waqar-ukaea: Could we handle MFEM consistently with the other mesh backends here? In particular, the `cmake/XDGConfig.cmake.in` should conditionally call `find_dependency(MFEM)` so that installed consumers can resolve the dependency and I think we mark MFEM PRIVATE here in the `CMakeLists.txt`

tests/CMakeLists.txt
Waqar-ukaea: I think I have an idea for this so I will just go ahead and try to implement it myself

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+  int num_faces() const override {
+    // first, look up the geom type
+    auto geom = mesh_manager_->get_volume_element_type(element_);
+
+    switch (geom) {
+      case VolumeElementType::TET: return 4;
+      case VolumeElementType::HEX: return 6;
+      default: fatal_error("");
+    }
+  }
```
Waqar-ukaea: `get_volume_element_type(MeshID volume)` is expecting a volume MeshID as its argument. It will compile because we use the same MeshID type between id spaces but technically we are querying a single element rather than topological volumes so we are passing the wrong ID space here. 

I think a simpler way of returning the correct number of faces would be via the `faces_` array which gets populated earlier anyway: 


```suggestion
  int num_faces() const override {
    return faces_.Size();
  }
```

I've also added a suggestion for a new test in `test_mfem.cpp` to exercise this.

tests/test_mfem.cpp
```diff
@@ -0,0 +1,432 @@
...
+  Position end {0.0, 0.0, 1000.0};
+  auto tracks = xdg->segments(volume, start, end);
+
+  double length = std::accumulate(tracks.begin(), tracks.end(), 0.0, [](double sum, const auto& track) {
+    return sum + track.second;
+  });
+
+  REQUIRE_THAT(length, Catch::Matchers::WithinAbs(10.0, 1e-6));
+}
+
```
Waqar-ukaea: Suggested extra test for the MFEM element accessor to ensure the suggested change [in this other comment](https://github.com/xdg-org/xdg/pull/248/changes#r4081875796) is correct. This could also be extended later on to test for quads/hexes

```suggestion

TEST_CASE("MFEM element accessor")
{
  auto mesh_manager = std::make_shared<MfemMeshManager>();
  mesh_manager->load_file("single-tet.exo");
  mesh_manager->init();

  auto elements = mesh_manager->get_volume_elements(1);
  REQUIRE(elements.size() == 1);

  auto accessor =
      ElementFaceAccessor::create(mesh_manager.get(), elements.front());

  REQUIRE(accessor->num_faces() == 4);

  for (int face = 0; face < accessor->num_faces(); ++face) {
    REQUIRE(accessor->face_vertices(face).size() == 3);
  }
}

```

tests/util.h
Waqar-ukaea: We need to add the equivalent interface for MFEM too: 


```suggestion
MFEM_Interface = std::integral_constant<MeshLibrary, MeshLibary::MFEM>;

```

src/mfem/mesh_manager.cpp
```diff
@@ -0,0 +1,431 @@
...
+
+    // TODO: does this return the same number for every element on
+    // sideset 3?
+
+    int volume = mesh_->GetAttribute(elem_no);
+
+    int face_no = mesh_->GetBdrElementFaceIndex(face_index);
+    // check if the connectivity is still there
+    int e1, e2;
+    mesh_->GetFaceElements(face_no, &e1, &e2);
```
Waqar-ukaea: My thoughts aren't fully formed here so this may not be the most coherent (sorry).

So my understanding is that `MfemMeshManager::init()` groups all existing MFEM boundary elements by boundary attribute, with each attribute becoming an XDG surface. Here we iterate over those groups and use only the first boundary element to determine the parent-volume pair for the whole XDG surface. Is it guaranteed that every boundary element in one attribute group has the same parent-volume pair?

Waqar-ukaea (changes requested):
This is a really solid start to the MFEM backend with a lot of the functionality already there. I have some concerns/questions around mapping between MFEM's internal ID spaces and how we map onto XDG's but I think a lot of that can be hashed out in further comments.

I'll admit I am a little bit confused on the concept of MFEM `attributes`. Are these unique IDs for each volume? And similarly there are boundary attributes for each surface. Is it guaranteed that each volume attribute maps uniquely to each topological volume in XDG and likewise each boundary attribute maps uniquely to each topological surface in XDG?

I have a few other general comments and change suggestions as well as some wider discussion points for how we probably want the MeshManager to behave moving forwards as we begin to support additional mesh backends. But those changes can probably come in separate PRs.

Generally, I am happy for this PR to be merged without full support for quads/hexes but you should open a new issue dictating what needs to be changed/added to bring in that support as a subsequent PR.

tests/test_mesh_library_cross_check.cpp
Waqar-ukaea: Opened a PR to handle this #252

tests/CMakeLists.txt
Waqar-ukaea: Opened #252 to handle this

.github/workflows/ci.yml
Waqar-ukaea: We can also add the MFEM build/clone to the openmc-test.yml. Although currently that workflow won't actually exercise any of the MFEM code pathways in XDG, it might be nice to have the setup there ready for when it may be useful

tests/CMakeLists.txt
Waqar-ukaea: Once MFEM is linked privately in the top level CMakeLists.txt (as per [comment](https://github.com/xdg-org/xdg/pull/248/changes#r3933315668), we will need to explicitly link MFEM privately here to the test executables in the same manner we do so for MOAB and libmesh.

include/xdg/mfem/mesh_manager.h
```diff
@@ -0,0 +1,292 @@
...
+    return surfaces_.size();
+  }
+
+  int num_ents_of_dimension(int dim) const override {
+    switch (dim) {
+      case 3: return num_volumes();
+      case 2: return num_surfaces();
+      default: return 0;
+    }
+  }
```
Waqar-ukaea: (Comment more for myself). Another method which can be pulled up into the `MeshManager` class. As far as I can tell MOAB requires it's own implementation but could also default to this too

src/mfem/mesh_manager.cpp
Waqar-ukaea: Ah, I see libmesh also enforces this same condition

PR #250, opened 2026-08-20
Element walk updates

pshriwise:
This PR improves the robustness of element walking such that random particle walks on a regularized mesh of tets or hexes match those of OpenMC's structured mesh representation. 

The primary updates here are:

  - a change to element point containment kernels connected to the search trees such that they are consistent with the raytracing operations performed during element walks. This ensures that the `find_element` operation agrees with raytracing in the initial element of a line segment laid onto the mesh, even when the original point lies on an element vertex or edge.
  - selection of the next element by both distance and face normal alignment. For rays that exit through edges or vertices of elements, face crossings with numerically equivalent distances are kept and a crossing is chosen for the face normal that is most aligned with the query ray's direction. This avoids glancing intersections with the next element.

New regression tests with hard-coded pathological tracks have been included here and added to a new set of regression tests in this module.

I'll be updating the XDG PR in OpenMC soon where direct comparisons of these representations are checked.


include/xdg/geometry/face_common.h
```diff
@@ -6,7 +6,8 @@
 
 namespace xdg {
 
-inline Direction triangle_normal(const std::array<Vertex, 3>& vertices) {
+template<typename T>
+inline Direction triangle_normal(const T& vertices) {
```
Waqar-ukaea, line 10: Not one for this PR but I imagine there are other places where we could benefit from switching over to template parameters for anything which requires stuff like vertices to be passed.

include/xdg/hexahedron_contain.h
```diff
@@ -16,8 +16,9 @@ namespace xdg {
  * using MBCN canonical Hex8 ordering (Tautges 2010).
  *
  * The hexahedron is defined by eight vertices ordered per the MBCN
- * canonical numbering. The containment test uses face-plane sign tests; points
- * on the boundary are considered inside.
+ * canonical numbering. The containment test casts toward the hexahedron
```
Waqar-ukaea: Should this say "casts rays toward the hexahedron centroid"?

include/xdg/geometry/plucker.h
```diff
@@ -160,6 +169,47 @@ inline PluckerIntersectionResult plucker_ray_tri_intersect(dp::vec3 vertices[3],
...
+//! \brief Determine if a ray intersects a triangle using Plücker coordinates
+//! \param triangle The triangle vertices
+//! \param origin The ray origin
+//! \param direction The ray direction
+//! \return True if the ray intersects the triangle, false otherwise
+template<typename T>
+inline bool plucker_line_intersects_triangle(const T& triangle,
+                                             const Position& origin,
+                                             const Direction& direction)
+{
```
Waqar-ukaea, line 180: Not sure how much of a concern this is, since GPRT development has taken a backseat but this header file is directly consumed by some of the slang files and `slangc` can't compile a c++ template which causes compilation of XDG to fail when `XDG_ENABLE_GPRT=ON`. 

Since this method doesn't actually need to be used anywhere in the slang code, as a quick fix, I suggest wrapping the method in a compile time block that stops `slangc` from trying to parse it. 

```cpp
#ifndef __SLANG__
//! \brief Determine if a ray intersects a triangle using Plücker coordinates
//! \param triangle The triangle vertices
//! \param origin The ray origin
//! \param direction The ray direction
//! \return True if the ray intersects the triangle, false otherwise
template<typename T>
inline bool plucker_line_intersects_triangle(const T& triangle,
                                             const Position& origin,
                                             const Direction& direction)
{
...
}
#endif \\ #ifndef __SLANG__
```

include/xdg/geometry/plucker.h
```diff
@@ -150,7 +159,7 @@ inline PluckerIntersectionResult plucker_ray_tri_intersect(dp::vec3 vertices[3],
   double v = plucker_coord0 * inverse_sum;
 
   // Barycentric coords check
-  if (u < 0.0 || v < 0.0 || (u + v) > 1.0) {
+  if (u < -PLUCKER_ZERO_TOL || v < -PLUCKER_ZERO_TOL || (u + v) > 1.0 + PLUCKER_ZERO_TOL) {
```
Waqar-ukaea:
```suggestion
  if (u < -dp::DBL_ZERO_TOL || v < -dp::DBL_ZERO_TOL || (u + v) > 1.0 + dp::DBL_ZERO_TOL) {
```

Similar to another comment in this file, this causes slang compilation to fail but this time because `PLUCKER_ZERO_TOL` isn't defined anywhere for the slang compilation path to see. I'm hesitant to bring in `constants.h` directly since that has a lot of C++ specific stuff that slang compilation will complain about. This method already makes use of the `dp` namespace which allows cross compilation between slang and c++ and that namespace has a constant `dp::DBL_ZERO_TOL` which maps to an equivalent definition as `PLUCKER_ZERO_TOL` during C++ compilation: 

```cpp
#ifdef __SLANG__

// Slang compilation, map dp::vec3 -> double3
namespace dp { 
  typedef double3 vec3;

  inline double dot(vec3 a, vec3 b) { return ::dot(a, b); }
  inline vec3 cross(vec3 a, vec3 b) { return ::cross(a, b); }
  inline double abs(double a) { return ::abs(a); }

  static const double DBL_ZERO_TOL = 20 * DBL_EPSILON;
  static const double INFTY = 1.7976931348623157e+308; // std::numeric_limits<double>::max() is not available in slang
}

#else
#include "xdg/vec3da.h"

// C++ compilation map dp::vec3 -> xdg::Vec3da
namespace dp {
  typedef xdg::Vec3da vec3;

  inline double dot(vec3 a, vec3 b) { return xdg::dot(a, b); }
  inline vec3 cross(vec3 a, vec3 b) { return xdg::cross(a, b); }
  inline double abs(double a) { return std::fabs(a); }

  static constexpr double DBL_ZERO_TOL = 20.0 * std::numeric_limits<double>::epsilon();
  constexpr double INFTY {std::numeric_limits<double>::max()};

}

#endif // #ifndef __slang__
```
I've tested locally with the suggestion above in place and no tests are failing.

src/mesh_manager_interface.cpp
```diff
@@ -117,15 +125,20 @@ MeshManager::walk_elements(MeshID starting_element,
...
   while (distance > 0) {
-    // find the exit point from the current element and determine the next element
-    // if one exists
+    // find the exit point from the current element and determine the next
+    // element if one exists
     auto exit = next_element(elem, r, u);
+    if (exit.second == INFTY && distance > 0) {
+      warning(fmt::format("Unable to find finite exit point from element {}. Exiting walk_elements with {} distance unaccounted for.", elem, distance));
+      break;
+    }
```
Waqar-ukaea, line 135: Should this be a `fatal_error()` rather than a warning? Do we expect the downstream app to do their own error handling in this instance?

src/mesh_manager_interface.cpp
```diff
@@ -150,32 +163,31 @@ MeshManager::walk_elements(MeshID starting_element,
...
-                           const Position& u) const
+                          const Position& r,
+                          const Position& u) const
 {
+  struct FaceCandidate {
+    MeshID element {ID_NONE};
+    double distance;
+    int face {ID_NONE};
+    double exiting_dot;
+  };
```
Waqar-ukaea, line 174: Looks as though the member `int face` is never actually used anywhere? Maybe worth dropping it?

src/mesh_manager_interface.cpp
```diff
@@ -186,52 +198,83 @@ MeshManager::next_element(MeshID current_element,
...
+      PluckerIntersectionResult result0, result1;
+      double exiting_dot0 = u.dot(triangle_normal(tri0));
+      if (exiting_dot0 >= 0.0) {
+      result0 = plucker_ray_tri_intersect(tri0.data(),
+                                                r,
+                                                u,
+                                                INFTY,
+                                                -1e-10,
+                                                false,
+                                                0);
```
Waqar-ukaea: Being pedantic here about indentation but can we align with the first parameter:
```suggestion
      result0 = plucker_ray_tri_intersect(tri0.data(),
                                          r,
                                          u,
                                          INFTY,
                                          -1e-10,
                                          false,
                                          0);
```

src/mesh_manager_interface.cpp
```diff
@@ -186,52 +198,83 @@ MeshManager::next_element(MeshID current_element,
...
+
+      double exiting_dot1 = u.dot(triangle_normal(tri1));
+      if (exiting_dot1 >= 0.0) {
+      result1 = plucker_ray_tri_intersect(tri1.data(),
+                                                r,
+                                                u,
+                                                INFTY,
+                                                -1e-10,
+                                                false,
+                                                0);
```
Waqar-ukaea: Being pedantic again...
```suggestion
      result1 = plucker_ray_tri_intersect(tri1.data(),
                                          r,
                                          u,
                                          INFTY,
                                          -1e-10,
                                          false,
                                          0);
```

src/mesh_manager_interface.cpp
```diff
@@ -186,52 +198,83 @@ MeshManager::next_element(MeshID current_element,
...
+  }
+
+  if (tied_candidates.size() == 1)
+    return {tied_candidates.front()->element, tied_candidates.front()->distance};
+
+  const auto selected = *std::max_element(
+    tied_candidates.begin(), tied_candidates.end(),
+    [](const auto a, const auto b) { return a->exiting_dot < b->exiting_dot; });
+
+  return {selected->element, selected->distance};
```
Waqar-ukaea, line 276: No suggestions here but just wanted to say, this is pretty neat!

tests/test_tally_segments.cpp
```diff
@@ -46,14 +46,13 @@ TEST_CASE("Test Walk Elements") {
...
     std::cout << "Remaining distance: " << distance << std::endl;
-    // if(distance <= TINY_BIT) break;
 
     std::cout << "---" << std::endl;
     std::cout << "Current element: " << element << std::endl;
     std::cout << r << std::endl;
     std::cout << u << std::endl;
     std::tie(next_element, exit_distance) = xdg->next_element(element, r, u);
-    REQUIRE(exit_distance < distance);
+    REQUIRE(exit_distance <= distance);
```
Waqar-ukaea, line 55: It's not immediately obvious to me, why the equality change here?

Waqar-ukaea (changes requested):
Generally looks pretty solid and I can see how this should result in some more robust next_element selection. The regression tests look good too. 

I've raised a couple of minor concerns about GPRT/slang compilation which should be easily resolved.

include/xdg/hexahedron_contain.h
pshriwise: Mmmm yep yep. Thanks!

include/xdg/geometry/face_common.h
pshriwise: True. I think we probably address them as needed like we did here. 👍

include/xdg/geometry/plucker.h
pshriwise: We only call this in one place, so there isn't really a need to have it templated. I'll update the signature to something more slang-friendly in a bit.

src/mesh_manager_interface.cpp
pshriwise: Yeah I think I'd prefer that the application code handle this if possible. At least I'd like it to until https://github.com/xdg-org/xdg/issues/180 is addressed. I'll add a TODO comment here as a reminder to potentially change this once error handling can be caught in different ways by the application code depending on XDG configuration settings.

src/mesh_manager_interface.cpp
pshriwise: Good point! Removing...

src/mesh_manager_interface.cpp
pshriwise: On board! Should be updated now.

tests/test_tally_segments.cpp
pshriwise: The original equality was overly restrictive. It should be valid to return a sum that is exactly the distance passed into `next_element`.

include/xdg/geometry/plucker.h
pshriwise: I took a crack at this. Let me know what you think!

pshriwise:
Thanks for highlighting some of the slang incompatibilities here @Waqar-ukaea. I took a crack at fixing those and addressed your other comments. 

The latest commit plays around with a start for supporting cross-compatible templated functions that work for both C++/slang. I started adjusting types instead, but at some point we hit redundant memory copies to get higher-level functions to work with the signatures in `plucker.h`. As long as we keep it contained to those common intersection functions and avoid anything too fancy in the signatures I think it's manageable, but let me know what you think! I can try to extend it to the `plucker_ray_tri_intersect` function.

pshriwise:
It became pretty clear that this wasn't the place to take on common templating for C++ and slang. Doing it for just that method where the templating isn't really exercised doesn't make sense and modifying existing templates is going to muddy the scope of this PR. I've made some updates to hopefully remove some of the redundant copy/allocation in the methods that are in line with the way things are handled for tet containment. Maybe we can revisit this later.

Waqar-ukaea:
> It became pretty clear that this wasn't the place to take on common templating for C++ and slang. Doing it for just that method where the templating isn't really exercised doesn't make sense and modifying existing templates is going to muddy the scope of this PR. 

Agreed on it being out of scope. But I took a look at the commit and that seems like an interesting idea to go between C++ templates and slang generics but I couldn't get it compile the slang side of things. I took a look at the slang docs about generics and I think whilst they fill the same niche of providing compile time polymorphism I think their underlying implementations are quite different. From https://shader-slang.org/docs/understanding-generics/ : 


image

So I think the easiest way of ensuring cross-compatability will likely be having a common "core" C method that can be called by both a C++ template and a slang generic

src/mesh_manager_interface.cpp
```diff
@@ -117,15 +125,21 @@ MeshManager::walk_elements(MeshID starting_element,
...
-    // find the exit point from the current element and determine the next element
-    // if one exists
+    // find the exit point from the current element and determine the next
+    // element if one exists
     auto exit = next_element(elem, r, u);
+    if (exit.second == INFTY && distance > 0) {
+      // TODO: Update to fatal_error once configurable error handling is implemented
+      warning(fmt::format("Unable to find finite exit point from element {}. Exiting walk_elements with {} distance unaccounted for.", elem, distance));
+      break;
+    }
```
Waqar-ukaea, line 135: Codex suggested a potential infinite loop bug when `walk_elements()` is called via the overload where a specific volume is not given:
```cpp
std::vector<std::pair<MeshID, double>>
XDG::segments(const Position& start,
              const Position& end) const
```
In that function we loop:
```cpp 
while (distance > 0) {
   find_element(r, ...);
   ray_fire(r, ...);
   ... 
   walk_elements(r, ...);
   r += u * segment_sum; // so if segment_sum = 0, then r doesn't change
   distance -= segment_sum; // so if segment_sum = 0, then distance doesn't reduce
}
```
Essentially if `walk_elements()` fails to find an exit on it's first iteration it would return an empty vector right? Which would result in `segment_sum == 0.0` so the outer distance reduction loop never gets smaller and the position being queried never changes, so we could end up finding the same element, going into `walk_elements()` and it keeps doing that loop. 

Maybe this won't ever actually happen in practice though but the code does leave room for it i guess, so i wonder if an empty vector return from `walk_elements()` should be handled explicitly?

Waqar-ukaea (commented):
One last comment but otherwise happy to merge once addressed

src/mesh_manager_interface.cpp
pshriwise: Hrmm yeah I don't think we actually exercise that overload in OpenMC, just in the tests here. All the same I think we should address it. I'll take a look. Thanks!

pshriwise: Summarizing online convo from yesterday: 

Looking more at this, I think it's an inherent risk in this particular method. At first glance making an adjustment when the number of segments returned is zero seems okay, but we could also get a single segment with a zero distance intersection in theory. That leaves us to making floating point adjustments when `segment_sum` is zero, and without sufficient pathological cases we'd like to avoid making such a change without cause.

